PR #33, opened 2023-10-29
Fix : Add anthropic version header with each request

insaaniManav:
The anthropic implementation was breaking so added a version header by default , we could potentially make this configurable but there is no publically available list of versions as per my research


VisargD:
Hey @insaaniManav . The version that you added as default now supports incremental chunk streaming. But the implementation in Rubeus does the old stream where each chunk is a combination of all the old chunks + the new one. Let me sit on it today and fix it.

VisargD:
Closing this PR as this requires some more changes than just hardcoding the version header. Raised a new PR #38

PR #42, opened 2023-11-07
fix: adding mappings for max_tokens and stopsequences

### LINKED ISSUE #43: Fix: Adding the mappings for max_tokens and sequences in PaLM
The PaLM APIs were missing the mappings for max_tokens and sequences and as a result they were not being picked up.

noble-varghese:
Fixes: #43

### DIFF
--- a/src/providers/palm/chatComplete.ts
+++ b/src/providers/palm/chatComplete.ts
@@ -50,6 +50,15 @@ export const PalmChatCompleteConfig: ProviderConfig = {
         default: 1,
         min: 1,
         max: 8
+    },
+    max_tokens: {
+        param: "maxOutputTokens",
+        default: 100,
+        min: 1,
+    },
+    stop: {
+        param: "stopSequences"
+
     }
 };
 
--- a/src/providers/palm/complete.ts
+++ b/src/providers/palm/complete.ts
@@ -44,7 +44,17 @@ export const PalmCompleteConfig: ProviderConfig = {
         default: 1,
         min: 1,
         max: 8
+    },
+    max_tokens: {
+        param: "maxOutputTokens",
+        default: 100,
+        min: 1,
+    },
+    stop: {
+        param: "stopSequences"
+
     }
+
 };
 
 export const PalmCompleteResponseTransform: (response: PalmCompleteResponse, responseStatus: number) => CompletionResponse | ErrorResponse = (response, responseStatus) => {


VisargD (approved):

wanguardd (changes requested):
System of Review completed analysis of 'fix: adding mappings for max_tokens and stopsequences'. Found **5 findings** across 2 files.

## 📋 Review Summary

| Category | Count |
|----------|-------|
| 🐛 CORRECTNESS | 2 |
| 🔍 QUALITY | 2 |
| ✅ TESTING | 1 |

---

## 🐛 CORRECTNESS

**`stop` parameter silently passes a bare string to Palm API, which requires an array — stop-sequence functionality is broken for all string inputs**

The gateway's `stop` field accepts `string | string[]` (per `src/types/requestBody.ts:442`). Palm's `stopSequences` field requires `string[]`. The PR adds the mapping but omits the `transform` function that converts between the two. The transformation pipeline in `src/services/transformToProviderRequest.ts:37` only applies a conversion when `paramConfig.transform` is defined — without it, `getValue()` returns the raw input unchanged, and `stopSequences: "."` (a string) gets sent to Palm instead of `stopSequences: ["."]`. The Palm API will reject or ignore a non-array value, silently discarding the stop sequence.

**Evidence:**
```typescript
// src/services/transformToProviderRequest.ts:37
if (paramConfig.transform) {
  value = paramConfig.transform(params, providerOptions);
}
// No transform → raw string passes through unchanged

// src/types/requestBody.ts:442
stop?: string | string[];

// PR diff — no transform field:
stop: {
    param: "stopSequences"   // ← no transform
}
```

The Cohere provider (same codebase) handles this correctly at `src/providers/cohere/chatComplete.ts:49-58`:
```typescript
stop: {
  param: 'stop_sequences',
  transform: (params: Params) => {
    if (typeof params.stop === 'string') {
      return [params.stop];
    }
    return params.stop;
  },
},
```

**Action:**
- [ ] Add the same `transform` to `stop` in both `src/providers/palm/chatComplete.ts` and `src/providers/palm/complete.ts`:
  ```typescript
  stop: {
    param: 'stopSequences',
    transform: (params: Params) =>
      typeof params.stop === 'string' ? [params.stop] : params.stop,
  },
  ```

---

**Response transforms hardcode `finish_reason: 'length'` for all completions, which is now misleading given stop-sequence support is being introduced**

`src/providers/palm/chatComplete.ts:102` and `src/providers/palm/complete.ts:88` both emit `finish_reason: 'length'` unconditionally. Once stop sequences work correctly (after the fix above), callers will receive `finish_reason: 'length'` even when termination was caused by a stop sequence — the opposite of what the field is supposed to communicate. The OpenAI spec uses `'stop'` for normal completion and stop-sequence termination, and `'length'` only when the max-token limit was hit.

**Evidence:**
```typescript
// src/providers/palm/chatComplete.ts:99-103
choices:
  response.candidates?.map((generation, index) => ({
    message: { role: 'assistant', content: generation.content },
    index: index,
    finish_reason: 'length',   // ← always 'length', regardless of how generation stopped
  })) ?? [],
```

**Action:**
- [ ] Change the default `finish_reason` to `'stop'` in both response transforms. If the Palm API response provides a signal for max-token truncation, map that to `'length'` explicitly; otherwise `'stop'` is the correct default for this API.

---

## 🔍 QUALITY

**Double quotes used for string literals, inconsistent with the codebase-wide single-quote convention**

Every string literal in the existing provider configs uses single quotes. The PR introduces double-quoted values in the new entries.

**Evidence:**
```diff
+    max_tokens: {
+        param: "maxOutputTokens",   ← double quotes
+    },
+    stop: {
+        param: "stopSequences"      ← double quotes
+    }
```
Existing entries in the same file:
```typescript
model: { param: 'model', required: true, default: 'model/chat-bison-001' },
temperature: { param: 'temperature', default: 1, min: 0, max: 1 },
```

**Action:**
- [ ] Replace `"maxOutputTokens"` with `'maxOutputTokens'` and `"stopSequences"` with `'stopSequences'` in both files

---

**Spurious blank lines inside and after the `stop` config object**

**Evidence:**
```diff
+    stop: {
+        param: "stopSequences"
+                                  ← blank line inside braces
     }
```
In `complete.ts`, there's also an extra blank line after the `stop` block before the closing `};`. No other `ParameterConfig` entry in this file or any other provider uses blank lines inside config objects.

**Action:**
- [ ] Remove the blank line between `param: "stopSequences"` and the closing brace in `chatComplete.ts`
- [ ] Remove the extra blank line after the `stop` block in `complete.ts`

---

## ✅ TESTING

**No tests added for 19 lines of new production code; the string→array type-mismatch bug went undetected due to absence of test coverage for the Palm provider**

The PR adds 19 lines (+19 additions, 0 deletions per GitHub). A search of the entire `tests/` directory finds zero files referencing Palm configs, `PalmChatComplete`, `PalmComplete`, or palm-related transforms. The `transformUsingProviderConfig` function, which executes the ParameterConfig pipeline for all providers, also lacks standalone tests. Had there been a test for `stop: "."` → `stopSequences: ["."]`, the missing transform would have been caught immediately.

**Evidence:**
```bash
find tests/ -name "*.ts" | xargs grep -l "palm\|PalmConfig\|PalmChat\|PalmComplete"
# Returns: (no output)
```
PR additions: 19 lines (via `gh pr view 42 --json additions`). Palm test files: 0.

**Action:**
- [ ] Add unit tests for the Palm provider covering at minimum:
  1. `stop: "."` → `stopSequences: ["."]` (string input converted to array, verifying the transform)
  2. `stop: [".", "!"]` → `stopSequences: [".", "!"]` (array input passes through unchanged)
  3. `max_tokens: 50` uses the `maxOutputTokens` param name in the transformed request
  4. `max_tokens` omitted → default of 100 is applied
  5. `PalmChatCompleteResponseTransform` maps `candidates` to the correct `choices` structure

---

The `max_tokens → maxOutputTokens` mapping is correct and the basic structure of the PR is on the right track. The critical issue is the missing `transform` on `stop`, which makes the feature non-functional for the most common usage pattern (string stop values). Addressing Finding 1 is the blocker.

---

❗ > **Note**: This System of Review operates with the same information access and constraints as any human developer working on this repository. While the system is highly accurate (99%+ reliability), any confusion or apparent misunderstanding typically indicates gaps in repository documentation, unclear code organization, or insufficient project discipline - the same issues that would impact any team member. If the System cannot locate critical information and draws incorrect conclusions, a human developer would face identical challenges. We maintain high standards to respect our teammates' time, our clients' investment, and the integrity of this project. Please ensure the repository provides clear, discoverable context for all reviewers.
>
> **When addressing review feedback**: Open a separate commit for each point you address, clearly referencing the finding in the commit message. This ensures knowledge is captured in the repository history and helps other developers avoid the same pitfalls. If you cannot address a point, leave a detailed comment in this PR review thread explaining specifically what is wrong with the finding or why it cannot be addressed - never ignore feedback silently. Use comments only in rare cases when the System has genuinely missed existing context in the repository; prefer commits as the primary response mechanism to build institutional knowledge.

PR #149, opened 2024-01-17
feat: Add Deep Infra Provider

### LINKED ISSUE #110: [Provider] Add Deepinfra
Docs: https://deepinfra.com/docs

Suraj-Bhandarkar-S:
**Title: Add Deep Infra** 

- Add provider as Deep Infra
- Support /chat/completions route

Made the addition of Deep Infra provider using the docs provided for reference, followed the structure from other already existing API. Let me know if I need to make changes or if I have missed any variables. 

- #110

### DIFF
--- a/src/providers/deepinfra/api.ts
+++ b/src/providers/deepinfra/api.ts
@@ -5,7 +5,6 @@ const DeepInfraApiConfig: ProviderAPIConfig = {
   headers: (API_KEY: string) => {
     return {
       Authorization: `Bearer ${API_KEY}`,
-      "Content-Type": "application/json",
     };
   },
   chatComplete: "/chat/completions",
--- a/src/providers/deepinfra/api.ts
+++ b/src/providers/deepinfra/api.ts
@@ -1,0 +1,14 @@
+import { ProviderAPIConfig } from "../types";
+
+const DeepInfraApiConfig: ProviderAPIConfig = {
+  baseURL: "https://api.deepinfra.com/v1/openai",
+  headers: (API_KEY: string) => {
+    return {
+      Authorization: `Bearer ${API_KEY}`,
+      "Content-Type": "application/json",
+    };
+  },
+  chatComplete: "/chat/completions",
+};
+
+export default DeepInfraApiConfig;
--- a/src/providers/deepinfra/chatComplete.ts
+++ b/src/providers/deepinfra/chatComplete.ts
@@ -108,14 +108,23 @@ export const DeepInfraChatCompleteResponseTransform: (
     responseStatus !== 200 &&
     response.detail.length
   ) {
-    const firstError = response.detail[0];
-    const errorField = firstError.loc.join(".");
-    const errorMessage = firstError.msg;
-    const errorType = firstError.type;
+    let firstError: Record<string, any> | undefined;
+    let errorField: string | null = null;
+    let errorMessage: string | undefined;
+    let errorType: string | null = null;
+
+    if (Array.isArray(response.detail)) {
+      [firstError] = response.detail;
+      errorField = firstError?.loc?.join(".") ?? "";
+      errorMessage = firstError.msg;
+      errorType = firstError.type;
+    } else {
+      errorMessage = response.detail;
+    }
 
     return {
       error: {
-        message: `${errorField}: ${errorMessage}`,
+        message: `${errorField ? `${errorField}: ` : ""}${errorMessage}`,
         type: errorType,
         param: null,
         code: null,
--- a/src/providers/deepinfra/chatComplete.ts
+++ b/src/providers/deepinfra/chatComplete.ts
@@ -1,0 +1,155 @@
+import { DEEPINFRA } from "../../globals";
+import {
+  ChatCompletionResponse,
+  ErrorResponse,
+  ProviderConfig,
+} from "../types";
+
+export const DeepInfraChatCompleteConfig: ProviderConfig = {
+  model: {
+    param: "model",
+    required: true,
+    default: "meta-llama/Llama-2-70b-chat-hf",
+  },
+  messages: {
+    param: "messages",
+    default: [],
+  },
+  temperature: {
+    param: "temperature",
+    default: 0.7,
+    min: 0,
+    max: 1,
+  },
+  top_p: {
+    param: "top_p",
+    default: 1,
+    min: 0,
+    max: 1,
+  },
+  max_tokens: {
+    param: "max_tokens",
+    default: null,
+    min: 1,
+  },
+  stream: {
+    param: "stream",
+    default: false,
+  },
+};
+
+interface DeepInfraChatCompleteResponse extends ChatCompletionResponse {
+  id: string;
+  object: string;
+  created: number;
+  model: string;
+  usage: {
+    prompt_tokens: number;
+    completion_tokens: number;
+    total_tokens: number;
+  };
+}
+
+export interface DeepInfraErrorResponse {
+    object: string;
+    message: string;
+    type: string;
+    param: string | null;
+    code: string;
+}
+
+interface DeepInfraStreamChunk {
+    id: string;
+    object: string;
+    created: number;
+    model: string;
+    choices: {
+        delta: {
+            role?: string | null;
+            content?: string;
+        };
+        index: number;
+        finish_reason: string | null;
+    }[];
+}
+
+export const DeepInfraChatCompleteResponseTransform: (
+  response: DeepInfraChatCompleteResponse | DeepInfraErrorResponse,
+  responseStatus: number
+) => ChatCompletionResponse | ErrorResponse = (response, responseStatus) => {
+    if ("message" in response && responseStatus !== 200) {
+        return {
+            error: {
+                message: response.message,
+                type: response.type,
+                param: response.param,
+                code: response.code,
+            },
+            provider: DEEPINFRA,
+        } as ErrorResponse;
+    }
+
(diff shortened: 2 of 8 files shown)


VisargD:
Hey @Suraj-Bhandarkar-S . Thanks for the PR. Will do a detailed review and also run tests for these once. Looks good at first glance.

src/providers/deepinfra/chatComplete.ts
```diff
@@ -0,0 +1,151 @@
+import { DEEPINFRA } from "../../globals";
+import {
+  ChatCompletionResponse,
+  ErrorResponse,
+  ProviderConfig,
+} from "../types";
+
+export const DeepInfraChatCompleteConfig: ProviderConfig = {
```
VisargD, line 11: Can you please set the default, min and max values based on OpenAI docs.
https://platform.openai.com/docs/api-reference/chat/create

VisargD: The model default can still be of DeepInfra. If they do not have a default model mentioned, then we can pick up their most popular one or the one that they use in examples in their docs

Suraj-Bhandarkar-S: @VisargD So, I was not able to find specific default model saw deepinfra/airoboros-70b in their models section can we use this? also [https://platform.openai.com/docs/api-reference/chat/create](https://platform.openai.com/docs/api-reference/chat/create)  has multiple fields but I have used the below fields as this docs had mentioned and the default values have been used from the above mentioned url. If I need to make any changes let me know :) Thanks.
image

Suraj-Bhandarkar-S: ++update there was few endpoints using additional param's as per docs so add those as well.

VisargD:
I have found one issue. You have not added deepinfra in this file: `src/providers/index.ts`. Without that, the integration will not work. Can you please add that. Also, please make sure that all the flows are tested on your end.

src/providers/deepinfra/index.ts
```diff
@@ -0,0 +1,14 @@
+import { ProviderConfigs } from "../types";
+import DeepInfraApiConfig from "./api";
+import {DeepInfraChatCompleteConfig, DeepInfraChatCompleteStreamChunkTransform} from "./chatComplete";
+
+const DeepInfraConfig: ProviderConfigs = {
+    chatComplete: DeepInfraChatCompleteConfig,
+    api: DeepInfraApiConfig,
+    responseTransforms: {
+        chatComplete: DeepInfraChatCompleteConfig,
```
VisargD: Looks like this is incorrect. The transform function needs to be used here instead of the Config object. Can you please fix this.

src/providers/deepinfra/chatComplete.ts
```diff
@@ -0,0 +1,176 @@
...
+    delta: {
+      role?: string | null;
+      content?: string;
+    };
+    index: number;
+    finish_reason: string | null;
+  }[];
+}
+
+export const DeepInfraChatCompleteResponseTransform: (
```
VisargD, line 102: Error structure for deepinfra is different than OpenAI. You can replicate the error message by sending invalid fields to deepinfra like max_tokens="abc" or something. It sends an array of failed validation like this

```json
{
    "detail": [
        {
            "loc": [
                "body",
                "stream"
            ],
            "msg": "value could not be parsed to a boolean",
            "type": "type_error.bool"
        },
        {
            "loc": [
                "body",
                "max_tokens"
            ],
            "msg": "value is not a valid integer",
            "type": "type_error.integer"
        }
    ]
}
```

In this function instead of checking `if ("message" in response && responseStatus !== 200)`, we can check for `if ("detail" in response && responseStatus !== 200 && response.detail.length)` and then map the first error from the detail array to gateway error structure. Something like this:

```
const firstError = response.detail[0];
const errorField = firstError.loc.join(".");
const errorMessage = firstError.msg
const errorType = firstError.type

return {
  error: {
    message: `${errorField}: ${errorMessage}`,
    type: errorType,
    param: null,
    code: null,
  },
  provider: DEEPINFRA,
} as ErrorResponse;
```

VisargD:
The split pattern used by deepinfra between each stream chunk is `\r\n\r\n`. You can check this by printing JSON.stringify() of stream chunks that come in response from deepinfra. This pattern needs to be defined in `src/utils.ts` in this function
https://github.com/Portkey-AI/gateway/blob/b5dbfadd5bc697c5c212cf44a3bc8bea898f3216/src/utils.ts#L4

Without this, gateway is not able to recognize when a particular stream chunk ends.

src/providers/deepinfra/chatComplete.ts
Suraj-Bhandarkar-S: Thank You for guiding me!! will check it out

Suraj-Bhandarkar-S:
@VisargD when I am trying to test in local for deepinfra provider,
> ```
>`curl --location '127.0.0.1:8787/v1/openai/chat/completions' \
>--header 'x-portkey-provider: deepinfra' \
>--header 'Authorization: Bearer (added my key)' \
>--header 'Content-Type: application/json' \
>--data '{"messages": [{"role": "user","content": "Say this is test."}], "max_tokens": 20, "model": "deepinfra/airoboros-70b"}'`

I am getting 
> ```
>{
>   "status": "failure",
>   "message": "Invalid provider passed"
>}

I have added the provider as you had mentioned I am not sure if I have missed anything. (Using PostMan to Test) Thanks again!

src/providers/deepinfra/api.ts
```diff
@@ -0,0 +1,14 @@
+import { ProviderAPIConfig } from "../types";
+
+const DeepInfraApiConfig: ProviderAPIConfig = {
+  baseURL: "https://api.deepinfra.com/v1/openai",
+  headers: (API_KEY: string) => {
+    return {
+      Authorization: `Bearer ${API_KEY}`,
+      "Content-Type": "application/json",
```
VisargD: We already add content-type header in the below function. So we can remove it from here. Otherwise, this will cause duplicate header
https://github.com/Portkey-AI/gateway/blob/b5dbfadd5bc697c5c212cf44a3bc8bea898f3216/src/handlers/handlerUtils.ts#L20-L23

VisargD:
> when I am trying to test in local for deepinfra provider,
> curl --location '127.0.0.1:8787/v1/openai/chat/completions' \ --header 'x-portkey-provider: deepinfra' \ --header 'Authorization: Bearer (added my key)' \ --header 'Content-Type: application/json' \ --data '{"messages": [{"role": "user","content": "Say this is test."}], "max_tokens": 20, "model": "deepinfra/airoboros-70b"}'

- The url needs to be `http://127.0.0.1:8787/v1/chat/completions` . You do not have to send openai in the url as it is picked up from the config itself.
- Also looks like the formatting of the curl is wrong. Here is a working version
```
curl --location 'http://127.0.0.1:8787/v1/chat/completions' \
--header 'x-portkey-provider: deepinfra' \
--header 'Authorization: Bearer (added my key)' \
--data '{"messages": [{"role": "user","content": "Say this is test."}], "max_tokens": 20, "model": "deepinfra/airoboros-70b"}'
```

src/providers/deepinfra/chatComplete.ts
VisargD: Sorry for opening this again. But I also found out that there is one more version of their error response.
```
{
    "detail": "User is not authorized to access this resource"
}
```

So we might have to put a typeof check in detail field. If its a string then we can just show it in message and keep everything as null. And if its an array, then we can use the current approach. 

You can replicate this error by sending wrong api key.

Suraj-Bhandarkar-S:
I tried the above still same not sure why, if I change to other provider like mistral-ai, openai at least the request is being hit error or unauthorized access.

VisargD:
It might be possible that you are on a different branch or you forgot to restart your local server after making the latest changes.

src/providers/deepinfra/chatComplete.ts
VisargD: Suggesting something like this:
```
    let firstError: Record<string, any> | undefined;
    let errorField: string | null = null;
    let errorMessage: string | undefined;
    let errorType: string | null = null;
    if (Array.isArray(response.detail)) {
        firstError = response.detail[0];
        errorField = firstError.loc?.join(".") ?? "";
        errorMessage = firstError.msg;
        errorType = firstError.type;
    } else {
        errorMessage = response.detail;
        errorField = null;
        errorType = null;
    }

    return {
      error: {
        message: `${errorField ? `${errorField}: ` : ''}${errorMessage}`,
        type: errorType,
        param: null,
        code: null,
      },
      provider: DEEPINFRA,
    } as ErrorResponse;
  }
  ```

Suraj-Bhandarkar-S:
> It might be possible that you are on a different branch or you forgot to restart your local server after making the latest changes.

I tried but still same issue, are you able to use the provider from your side ?

VisargD:
Yes. Try putting in some logs at some checkpoints to see whats happening. Also make sure that the port is correct (if you are using wrangler in local)

VisargD:
Please request a re-review once you are done with all the changes and testing

Suraj-Bhandarkar-S:
Was able to run the provider and test the requests in postman, I had to use wrangler to run in local. Thank you again :)

VisargD (approved):

VisargD:
LGTM

VisargD:
Closes #110

PR #194, opened 2024-02-06
Feat/ollama integration

### LINKED ISSUE #14: [Provider] Add Ollama models
Docs: https://github.com/jmorganca/ollama/blob/main/docs/api.md

### LINKED ISSUE #216: Add custom_host support to allow proxies
Problem statement:

Gateway currently does not allow hitting LLM which are under some other proxy because the base url for each provider is static in code. Many organisations or users have openai, anthropic, cohere, etc access under a proxy server. So instead of using api.openai.com, their request should be able to hit their custom proxy url to use that proxy. One of the main reasons of such proxy is to have a custom auth, spend limit, rate limit, etc. which organisations assign to their members instead of exposing the actual LLM key.

Related issue: https://github.com/Portkey-AI/gateway/issues/124

Solution: 
- add a custom_host setting in configs where users can pass their proxy url and gateway will just add the endpoint to it. 
- add a x-portkey-custom-host header which does the same as config setting but it will be applied globally and will overwrite custom_host passed in config.

### LINKED ISSUE #217: Add forward_headers support to allow passing headers to LLM directly
Add a setting named forward_headers which allows users to add header names which will be picked up from the request headers and will be directly passed to LLM request without any overwrite/manipulation from gateway. This will allow users to add custom auth headers which are not simple tokens but have multiple moving pieces  like session token, user token, IAM based token etc. 

This can be a part of configs as well as headers:
- forward_headers config setting which expects an array of header names.
- x-portkey-forward-headers header which accepts comma separated header names. This header will be applied globally for the request and will overwrite forward_headers sent in config for that call.

csgulati09:
**Title:** 
- Provider: Ollama

**Description:** (optional)
- Added Ollama as a provider
- Covering 3 major features: Complete, chat and embed
- Both, stream and non-stream functions are supported
- Add support for custom_host setting in config and alss x-portkey-custom-host header.
- Add support for forward_headers setting in config and also x-portkey-forward-headers headers.

**Motivation:** (optional)
- This allows the users to use their self hosted Ollama via Portkey gateway.
- To allow adding custom base url to use proxy instead of directly using provider base_url.
- To allow users to add custom headers to LLM request without any overwrite from gateway.

- Closes #14 
- Closes #217 
- Closes #216


src/providers/ollama/chatComplete.ts
```diff
@@ -0,0 +1,208 @@
...
+      created: parsedChunk.created,
+      model: parsedChunk.model,
+      provider: OLLAMA,
+      choices: parsedChunk.choices,
+    })}` + "\n\n"
+  );
+};
+
+// export interface OllamaChatCompleteResponse extends ChatCompletionResponse {
+//   system_fingerprint: string;
```
VisargD: Please remove all the comments which are not necessary

src/providers/ollama/api.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const OllamaAPIConfig: ProviderAPIConfig = {
+  headers: () => {
+    return null;
+  },
+  chatComplete: "/v1/chat/completions",
+  embed:"/api/embeddings",
+  getEndpoint: (fn: string, API_VERSION: string, url?: string) => {
+    let mappedFn = fn;
+    if (fn === "proxy" && url && url?.indexOf("/chat/completions") > -1) {
```
VisargD: This is not required as /chat/completions will always point to the unified route.

src/providers/ollama/api.ts
```diff
@@ -0,0 +1,28 @@
+import { ProviderAPIConfig } from "../types";
+
+const OllamaAPIConfig: ProviderAPIConfig = {
+  headers: () => {
+    return null;
+  },
+  chatComplete: "/v1/chat/completions",
+  embed:"/api/embeddings",
+  getEndpoint: (fn: string, API_VERSION: string, url?: string) => {
```
VisargD: Please remove API_VERSION as it is not used

src/providers/ollama/chatComplete.ts
```diff
@@ -0,0 +1,131 @@
+import {
+  ChatCompletionResponse,
+  CompletionResponse,
```
VisargD: nitpick: Please remove unused import

src/utils.ts
```diff
@@ -21,6 +21,9 @@ export const getStreamModeSplitPattern = (proxyProvider: string, requestURL: str
     {
       splitPattern = '\r\n\r\n';
     }
+    if(proxyProvider === OLLAMA){
```
VisargD: This can be removed as we already initialize splitPattern with `\n\n`

src/handlers/streamHandler.ts
```diff
@@ -99,8 +99,8 @@ export async function handleStreamingMode(response: Response, proxyProvider: str
         writer.close();
     })();
 
-    // Convert GEMINI/COHERE json stream to text/event-stream for non-proxy calls
-    if ([GOOGLE, COHERE].includes(proxyProvider) && responseTransformer) {
+    // Convert GEMINI/COHERE/OLLAMA json stream to text/event-stream for non-proxy calls
+    if ([GOOGLE, COHERE, OLLAMA].includes(proxyProvider) && responseTransformer) {
```
VisargD: This can be removed now as we have added support for OpenAI compatible ollama route. So the response from ollama will always be text/event stream

VisargD (approved):

PR #196, opened 2024-02-07
Adding a streaming run function for OpenAI Assistants

roh26it:
**Streaming `run` functions for Assistants API** 
- Created a new route for creating thread runs: `POST /threads/runs/stream` that automatically polls in the background and streams run steps and messages to the connected client.
- We capture the final run state and all the messages and runs in the log to be sent to Portkey

**Motivation:** (optional)
- One of the main reasons of low adoption of the Assistants API has been the unavailability of streaming. While this is still _faux_ streaming, it still makes the entire process a little more enjoyable and perceptibly fast.
- Capturing the right token usage and converting that to costs incurred is a happy side effect.


roh26it:
OpenAI supports streaming in their API now. Clearing this PR.

PR #246, opened 2024-03-06
fix: allow multiple user messages in a row for google gemini

### LINKED ISSUE #258: allow multiple user messages in a row for google gemini
Google does not allow multiple same-role messages in a row and it throws `Please ensure that multiturn requests ends with a user role or a function response.` error message.

This issue was created from this PR: https://github.com/Portkey-AI/gateway/pull/246 .

meronogbai:
**Title:** 
- This will allow users to include multiple user messages in a row in gemini-pro.

**Description:** (optional)
- Inspired by [this code](https://github.com/PublicAffairs/openai-gemini/blob/d16cfafa897364151dc4a4eac6d901e9b8c97e45/src/worker.mjs#L202-L215)

**Motivation:** (optional)
- An app I work on relies on multiple user messages and it was breaking for me when I tried to use it with gemini-pro.

**How to test?**

The following snippet breaks with gemini-pro even though it works with openai. Checkout this branch and everything should work without issues 🫡

```ts
async function main() {
  const stream = await openai.chat.completions.create({
    model: "gemini-pro",
    messages: ([
      {
        role: "user",
        content: "Say this is a test",
      },
      { role: "user", content: "now say im happy" },
    ]),
    stream: true,
    max_tokens: 4096,
  });

  for await (const chunk of stream) {
    process.stdout.write(chunk.choices[0]?.delta?.content || "");
  }
}

main();
```

### DIFF
--- a/src/providers/google/chatComplete.ts
+++ b/src/providers/google/chatComplete.ts
@@ -38,7 +38,10 @@ export const GoogleChatCompleteConfig: ProviderConfig = {
         param: "contents",
         default: "",
         transform: (params: Params) => {
-            const messages = params.messages?.map((message: Message) => {
+            let lastRole: 'user' | 'model' | undefined;
+            const messages: { role: string; parts: { text: string; }[]; }[] = [];
+
+             params.messages?.forEach((message: Message) => {
                 const role = message.role === "assistant" ? "model" : "user";
                 let parts = [];
                 if (typeof message.content === "string") {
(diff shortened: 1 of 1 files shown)


src/providers/google/chatComplete.ts
```diff
@@ -64,7 +67,17 @@ export const GoogleChatCompleteConfig: ProviderConfig = {
                         }
                     });
                 }
-                return { role, parts };
+
+                // @NOTE: This takes care of the "Please ensure that multiturn requests alternate between user and model."
+                // error that occurs when we have multiple user messages in a row.
+                const shouldAppendEmptyModeChat = lastRole === 'user' && role === 'user' && !params.model?.includes('vision');
```
meronogbai, line 73: Opting out in case of `gemini-pro-vision` because it doesn't support multiturn chat anyway.

vrushankportkey:
Thank you so much for the PR, @meronogbai! This is amazing. Paging @VisargD who can check this

VisargD:
Thanks for the PR @meronogbai . I will review and merge this PR today.

VisargD:
LGTM. Closes #258

VisargD (approved):

PR #268, opened 2024-03-24
Bedrock mistral

### LINKED ISSUE #274: Feat: Support Bedrock mistral models
Bedrock has now released mistral models. Gateway already supports other bedrock models and mistral can now be added.

Docs: https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-mistral.html

aashsach:
**Title:** 
- Support for Mistral models in bedorck

**Description:** (optional)
- Adding support for completion mistral bedrock

**Motivation:** (optional)
- Getting mistral response  from Portkey via bedrock

- #issue-number

### DIFF
--- a/src/providers/bedrock/complete.ts
+++ b/src/providers/bedrock/complete.ts
@@ -126,38 +126,38 @@ export const BedrockLLamaCompleteConfig: ProviderConfig = {
 };
 
 export const BedrockMistralCompleteConfig: ProviderConfig = {
-    prompt: {
-        param: "prompt",
-        required: true,
-    },
-    max_tokens: {
-        param: "max_tokens",
-        default: 20,
-        min: 1,
-    },
-    temperature: {
-        param: "temperature",
-        default: 0.75,
-        min: 0,
-        max: 5,
-    },
-    top_p: {
-        param: "p",
-        default: 0.75,
-        min: 0,
-        max: 1,
-    },
-    top_k: {
-        param: "k",
-        default: 0,
-        max: 500,
-    },
-    stop: {
-        param: "end_sequences",
-    },
-    stream: {
-        param: "stream",
-    },
+  prompt: {
+    param: 'prompt',
+    required: true,
+  },
+  max_tokens: {
+    param: 'max_tokens',
+    default: 20,
+    min: 1,
+  },
+  temperature: {
+    param: 'temperature',
+    default: 0.75,
+    min: 0,
+    max: 5,
+  },
+  top_p: {
+    param: 'p',
+    default: 0.75,
+    min: 0,
+    max: 1,
+  },
+  top_k: {
+    param: 'k',
+    default: 0,
+    max: 500,
+  },
+  stop: {
+    param: 'end_sequences',
+  },
+  stream: {
+    param: 'stream',
+  },
 };
 
 const transformTitanGenerationConfig = (params: Params) => {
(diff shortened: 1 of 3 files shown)


VisargD:
Hey @aashsach - Thanks for the PR. We have released some changes today related to code formatting as well as some changes related to provider config structure. You will have to update your branch with those changes.

VisargD:
Please let me know if you need any help with this.

src/providers/bedrock/complete.ts
```diff
@@ -754,3 +789,124 @@ export const BedrockCohereCompleteStreamChunkTransform: (
...
+
+export interface BedrockMistralCompleteResponse {
+  outputs: BedrockMistralCompleteOutput[];
+}
+
+export const BedrockMistralCompleteResponseTransform: (
+  response: BedrockMistralCompleteResponse | BedrockErrorResponse,
+  responseStatus: number
+) => CompletionResponse | ErrorResponse = (response, responseStatus) => {
+  if ('message' in response && responseStatus !== 200) {
```
VisargD: You can use `BedrockErrorResponseTransform` function to handle error for bedrock. 
https://github.com/Portkey-AI/gateway/blob/48785807bcc19dc13669354ac3bc895b9af58185/src/providers/bedrock/chatComplete.ts#L394-L405

Example usage: 
https://github.com/Portkey-AI/gateway/blob/48785807bcc19dc13669354ac3bc895b9af58185/src/providers/bedrock/complete.ts#L339-L344

src/providers/bedrock/complete.ts
```diff
@@ -754,3 +789,124 @@ export const BedrockCohereCompleteStreamChunkTransform: (
...
+      ],
+      usage: {
+        prompt_tokens: 0,
+        completion_tokens: 0,
+        total_tokens: 0,
+      },
+    };
+  }
+
+  return {
```
VisargD: Instead of writing the whole object, you can reuse an existing function:
https://github.com/Portkey-AI/gateway/blob/48785807bcc19dc13669354ac3bc895b9af58185/src/providers/bedrock/complete.ts#L370C1-L371C1

src/providers/bedrock/complete.ts
```diff
@@ -754,3 +789,124 @@ export const BedrockCohereCompleteStreamChunkTransform: (
...
+        provider: BEDROCK,
+        choices: [
+          {
+            text: parsedChunk.outputs[0].text,
+            index: 0,
+            logprobs: null,
+            finish_reason: parsedChunk.outputs[0].stop_reason,
+          },
+        ],
+        usage: {
```
VisargD, line 826: Bedrock sends token usage in the last chunk of the stream. I am not sure if that is the case for Mistral as well. Can you please check and  make the change accordingly. Here is an example: https://github.com/Portkey-AI/gateway/blob/48785807bcc19dc13669354ac3bc895b9af58185/src/providers/bedrock/complete.ts#L604-L612

VisargD:
@aashsach - For all the current bedrock models, we support chat as well as complete requests by making transformations accordingly. In case of mistral, we can transform `messages` into `prompt` and allow chat completions calls as well. You can check out the examples in `src/providers/bedrock/chatComplete.ts`. It has transformers written for many bedrock models. Will it be possible for you to add this change as well?

src/providers/bedrock/complete.ts
aashsach-oai: yup, you are right, they do send it in the last response. will update

src/providers/bedrock/complete.ts
aashsach: yup you are riight, they do send the response as well. will update

aashsach:
> @aashsach - For all the current bedrock models, we support chat as well as complete requests by making transformations accordingly. In case of mistral, we can transform `messages` into `prompt` and allow chat completions calls as well. You can check out the examples in `src/providers/bedrock/chatComplete.ts`. It has transformers written for many bedrock models. Will it be possible for you to add this change as well?

@VisargD mistral models are only text models, do not support chat modality as per
https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html

confirmed from their playground, chat playground doesn't support mistral

VisargD:
Correct. They are text models. But gateway internally makes transformations to convert chat completions input to text completions input. This is done to just make gateway more flexible in terms of request signature. You can check the `BedrockAI21ChatCompleteConfig` for an example

aashsach:
> Correct. They are text models. But gateway internally makes transformations to convert chat completions input to text completions input. This is done to just make gateway more flexible in terms of request signature. You can check the `BedrockAI21ChatCompleteConfig` for an example

done

src/providers/bedrock/complete.ts
```diff
@@ -125,6 +125,41 @@ export const BedrockLLamaCompleteConfig: ProviderConfig = {
...
+    default: 20,
+    min: 1,
+  },
+  temperature: {
+    param: 'temperature',
+    default: 0.75,
+    min: 0,
+    max: 5,
+  },
+  top_p: {
```
VisargD, line 144: As per their docs, this param is called top_p instead of p for mistral.
```
{
    "prompt": string,
    "max_tokens" : int,
    "stop" : [string],    
    "temperature": float,
    "top_p": float,
    "top_k": int
}
```

Docs: https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-mistral.html

VisargD: Same for top_k

src/providers/bedrock/complete.ts
```diff
@@ -125,6 +125,41 @@ export const BedrockLLamaCompleteConfig: ProviderConfig = {
...
+  },
+  top_k: {
+    param: 'k',
+    default: 0,
+    max: 200,
+  },
+  stop: {
+    param: 'end_sequences',
+  },
+  stream: {
```
VisargD: stream should not be a part of transform config as the request body does not accept it. The `getEndpoint` function will handle considering stream and adding it to url.

VisargD: Same for complete and chatComplete

src/providers/bedrock/complete.ts
aashsach: been testing it, seems to be working fine with p and k, will update as per doc regardless

VisargD: Somehow I get this error message for the `p` param
```
Malformed input request: #: extraneous key [p] is not permitted, please reformat your input and try again.
```

src/providers/bedrock/complete.ts
```diff
@@ -754,3 +785,115 @@ export const BedrockCohereCompleteStreamChunkTransform: (
...
+      provider: BEDROCK,
+      choices: [
+        {
+          text: response.outputs[0].text,
+          index: 0,
+          logprobs: null,
+          finish_reason: response.outputs[0].stop_reason,
+        },
+      ],
+      usage: {
```
VisargD, line 899: Can you please also add the tokens from headers as you did for chatComplete?

VisargD:
Apart from the one open comment related to complete non-stream token count, everything else looks good. I will run final tests and merge the PR once the comment is resolved.

src/providers/bedrock/complete.ts
```diff
@@ -125,6 +125,37 @@ export const BedrockLLamaCompleteConfig: ProviderConfig = {
...
+    param: 'top_p',
+    default: 0.75,
+    min: 0,
+    max: 1,
+  },
+  top_k: {
+    param: 'top_k',
+    default: 0,
+  },
+  stop: {
```
VisargD, line 154: As mentioned in this comment, stop can be mapped to stop itself.
https://github.com/Portkey-AI/gateway/pull/268/files#r1544562206

VisargD (approved):

VisargD:
LGTM

PR #345, opened 2024-05-09
Update the nodejs example

hargup:
- Added the required `max_tokens` parameter
- Add instructions on how to use portkey locally
- Printed the output to console

**Motivation:**
- The current example didn't work


vrushankportkey:
Thanks for the quick fix @hargup made one more change and pushed your suggestions to #346. Closing this PR now.

PR #363, opened 2024-05-14
bugfix: convert several providers stop_reason to valid openai finish_reason

unsync:
**Description:**
Currently, we forward the anthropic `stop_reason`, which is not compatible with the openai `finish_reason`.

This can cause warnings from some OpenAI sdk.

The conversion function returns `stop` as a default value in case new `stop_reason` are added.

There was also a bug in the groq and google providers for the `finish_reason`

**Side note:**
It would probably be beneficial to update the types to we can get compilation-time validation :  https://github.com/Portkey-AI/gateway/blob/77f32e57b14bd3585169b7bc884ff9ddf38e8ea1/src/providers/types.ts#L101-L123

Proposed change : 

```
export interface ChatChoice {
  index: number;
  message: Message;
  finish_reason: 'stop' | 'length' | 'function_call' | 'content_filter' | null;
  logprobs?: object | null;
}
```

```
export interface CompletionResponse extends CResponse {
  choices: {
    text: string;
    index: number;
    logprobs: null;
    finish_reason: 'stop' | 'length' | 'function_call' | 'content_filter' | null;
  }[];
}
```

see https://platform.openai.com/docs/guides/text-generation/chat-completions-response-format


unsync:
hello @vrushankportkey @VisargD, i'm not sure if there is a process after opening a PR to request a review, any chance this PR could be considered ?

vrushankportkey:
Thank you so much @unsync! Yes, we're checking the PR and will share review if any in some time! Really appreciate you pushing this bugfix 🙌

src/providers/anthropic/chatComplete.ts
```diff
@@ -345,6 +345,22 @@ export const AnthropicErrorResponseTransform: (
   return undefined;
 };
 
+// this converts the anthropic stop_reason to an openai finish_reason
+const getFinishReason = (stopReason?: string): string | null => {
```
narengogi, line 349: Recommend creating a type
```
type AnthropicStopReason = 'max_tokens' | 'stop_sequence' | 'end_turn' | ....
```

src/providers/anthropic/chatComplete.ts
```diff
@@ -345,6 +345,22 @@ export const AnthropicErrorResponseTransform: (
...
+  if (stopReason === 'max_tokens') {
+    return 'length';
+  }
+
+  if (stopReason === 'stop_sequence' || stopReason === 'end_turn') {
+    return 'stop';
+  }
+
+  // manages the case where the stop_reason is set but not recognized
+  return 'stop';
```
narengogi, line 361: This is okay, but not recommended, if we already know all the possible stop_reason supplied by anthropic, we should have a 1:1 mapping

src/providers/anthropic/chatComplete.ts
narengogi: also recommend using a switch/case

src/providers/anthropic/chatComplete.ts
```diff
@@ -394,7 +410,8 @@ export const AnthropicChatCompleteResponseTransform: (
           },
           index: 0,
           logprobs: null,
-          finish_reason: response.stop_reason,
+          finish_reason:
+            getFinishReason(response.stop_reason) || response.stop_reason,
```
narengogi, line 414: This should just be `getFinishReason(response.stop_reason)`, Otherwise it defeats the purpose of strictly typing the response

src/providers/google/chatComplete.ts
```diff
@@ -226,6 +226,28 @@ interface GoogleGenerateContentResponse {
   };
 }
 
+// this converts the google stop_reason to an openai finish_reason
+// https://ai.google.dev/api/python/google/ai/generativelanguage/Candidate/FinishReason
+const getFinishReason = (stopReason?: string): string | null => {
```
narengogi, line 354: Same comments as in the getFinishReason transformer function above.
1. Use switch/case
2. Define types for GoogleStopReason

src/providers/google/chatComplete.ts
```diff
@@ -291,7 +313,9 @@ export const GoogleChatCompleteResponseTransform: (
           return {
             message: message,
             index: generation.index,
-            finish_reason: generation.finishReason,
+            finish_reason:
+              getFinishReason(generation.finishReason) ??
```
narengogi, line 440: should only be
```
finish_reason:
           getFinishReason(generation.finishReason) 
```
default case should be handled inside `getFinishReason`

narengogi (changes requested):
Thanks for the Pull Request! Requested a few changes

narengogi:
@unsync I've created a fork of your branch to do some more thorough changes, it has your exisiting commits https://github.com/Portkey-AI/gateway/pull/468

unsync:
> @unsync I've created a fork of your branch to do some more thorough changes, it has your exisiting commits #468

@narengogi that is great !
I think your PR is a better candidate for merge than mine, i think my PR can be close in favour of yours ?

cc @vrushankportkey

vrushankportkey:
Thanks @unsync & @narengogi!

PR #373, opened 2024-05-21
Adding requestExecTime to requestOptions in the context

roh26it:
This improves response time metrics for streaming and cache use cases, without retries.


roh26it:
We can add metrics like time to first token also here.

VisargD:
Closing this in favour of another PR with a larger scope

PR #415, opened 2024-06-28
update cohere endpoints & response template

### LINKED ISSUE #321: Update Cohere to new endpoint
### What Happened?

I've just learned that Cohere deprecated their `generate` endpoint and that my fine-tune model is not working with Portkey. 

Ref: https://docs.cohere.com/reference/chat

How should we go about this? I am also open to make a PR but if there are some that are using, this may cause a breaking change. A simple major version bump would do?

_No response_

### Relevant Code Snippet

_No response_

### Your Twitter/LinkedIn

_No response_

b4s36t4:
**Title: Deprecate /generate endpoint for cohere provider** 
- Updates cohere endpoints in favour of depreciating  generate endpoint.

**Motivation:** (optional)

/closes #321


src/providers/cohere/complete.ts
VisargD, line 1: Hey. This is the complete handler so users would still be expected to send prompt instead of messages. I would suggest that we either ask for prompt and convert to messages with a transformer. Or else we can keep it same as the existing implementation and use /generate endpoint until cohere releases a final deprecation notice.

narengogi (changes requested):
I've checked out your branch and ran a test, but it fails.
Please find attached relevant curl request
```
--header 'x-portkey-provider: cohere' \
--header 'Authorization: $COHERE_API_KEY' \
--data '{
    "model": "command-r-plus",
    "max_tokens": 200,
    "stream": false,
    "messages": [
      {
        "role": "system",
        "content": "You are a helpful assistant"
    },
          {
        "role": "user",
        "content": "Hello Alfred, how are you?"
    }
    ]
}'
```

src/providers/cohere/chatComplete.ts
```diff
@@ -16,23 +15,39 @@ export const CohereChatCompleteConfig: ProviderConfig = {
...
+        const messages = params.messages || [];
+        const messagesWithoutLastMessage = messages.slice(
+          0,
+          messages.length - 1
+        );
+        // generate history and forware it to model
+        const history: (Message & { message?: string })[] =
+          messagesWithoutLastMessage.map((message) => ({
+            role: message.role,
+            message: message.content as string,
```
narengogi: also need to handle messages of object type
```
            {
        "role": "assistant",
        "content": [
          {
            "type": "text",
            "text": "is it me?"
          }
        ]
      }
```
You can refer https://platform.openai.com/docs/api-reference/chat/create for openai request structure.
Gateway transforms requests from OpenAI format to provider specific format, so all the possible request variations from. openai need to be handled by the gateway transformer

package-lock.json
VisargD, line 1: Can we remove this file from PR if its not required? I don't see any package.json changes

VisargD (approved):

PR #445, opened 2024-07-09
Triton Support

tapansstardog:
**Triton Support** 
- This branch contains the changes being done for Triton support. With these changes, portkey can now call Triton host. 

- The implementation is based on Ollama changes in the code.

**Description:**
Following is the sample configuration that we need to pass in order to call the Triton server running on `http://triton-host/v2/models/<model_name>`

```
{
  "strategy": {
      "mode": "fallback"
  },
  "targets": [
      {
        "provider": "triton",
        "custom_host": "http://triton-host/v2/models/<model_name>"
        "api_key": "empty"
      },
      {
        "provider": "together-ai",
        "model": "codellama/CodeLlama-34b-Instruct-hf",
         "api_key": "together_key"
      }
  ]
}
```
Sample Python code:

```
port_key = Portkey(
      api_key="empty", config=config, base_url="http://localhost:8787/v1"
)

completion: Union[
    TextCompletion, Iterator[TextCompletionChunk]
] = port_key.completions.create(prompt=prompt, **kwargs)

output = ""
if isinstance(completion, TextCompletion):
    output = completion.choices[0].text if completion.choices[0].text else ""
```

**Motivation:** 
- We wanted to have the fallback mechanism in place where we should connect to our own LLM hosted in our environment, `Triton` in our case. In case of failover, as a second option it should contact `TogetherAI`. Since Triton support was not there and we saw `Ollama` changes being recently pushed into PortKey, we thought of leveraging `custom_host` feature and made these changes.
- We have deployed TensorRT LLM on Triton server. With these changes we are able to achieve our requirements.

These changes do not have support for chatCompletions. We have tested our code with Text Completion use case where model repository of `TensorRT-LLM  backend `is hosted on `Triton`.

Thanks!

### DIFF
--- a/src/handlers/retryHandler.ts
+++ b/src/handlers/retryHandler.ts
@@ -119,8 +119,7 @@ export const retryRequest = async (
       lastResponse = new Response(error.message, {
         status: 503,
       });
-    }
-    else {
+    } else {
       lastResponse = new Response(error.message, {
         status: error.status,
         headers: error.headers,
--- a/src/handlers/retryHandler.ts
+++ b/src/handlers/retryHandler.ts
@@ -112,10 +112,20 @@ export const retryRequest = async (
       }
     );
   } catch (error: any) {
-    lastResponse = new Response(error.message, {
-      status: error.status,
-      headers: error.headers,
-    });
+    if (error instanceof TypeError && error.message?.includes('fetch failed')) {
+      console.error('Fetch failed:', error);
+      // This error comes in case the host address is unreachable. Empty status code used to get returned
+      // from here hence no retry logic used to get called.
+      lastResponse = new Response(error.message, {
+        status: 503,
+      });
+    }
+    else {
+      lastResponse = new Response(error.message, {
+        status: error.status,
+        headers: error.headers,
+      });
+    }
     console.warn(
       `Tried ${lastAttempt ?? 1} time(s) but failed. Error: ${JSON.stringify(error)}`
     );
--- a/src/handlers/retryHandler.ts
+++ b/src/handlers/retryHandler.ts
@@ -112,8 +112,12 @@ export const retryRequest = async (
       }
     );
   } catch (error: any) {
-    if (error instanceof TypeError && error.message?.includes('fetch failed')) {
-      console.error('Fetch failed:', error);
+    if (
+      error instanceof TypeError &&
+      error.cause instanceof Error &&
+      error.cause?.name === 'ConnectTimeoutError'
+    ) {
+      console.error('ConnectTimeoutError: ', error.cause);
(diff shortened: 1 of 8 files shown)


vrushankportkey:
Thank you so much for contributing, @tapansstardog! We are reviewing the PR now.

src/handlers/retryHandler.ts
```diff
@@ -109,10 +109,19 @@ export const retryRequest = async (
       }
     );
   } catch (error: any) {
-    lastResponse = new Response(error.message, {
-      status: error.status,
-      headers: error.headers,
-    });
+    if (error instanceof TypeError && error.message?.includes('fetch failed')) {
```
narengogi: This check should not be present in the retryHandler, the string is generic and this might get caught unintentionally

narengogi (changes requested):
I've no way to run and verify a triton inference server, but looks good apart from the comment

src/handlers/retryHandler.ts
narengogi: You can throw this from inside `TritonErrorResponseTransform`

tapansstardog: @narengogi I get the point. The error message handling was generic here. 

The error comes in case the user has provided incorrect IP Address of `custom_host` while we are making the connection with the server. At that time, the code doesn't reach till `TritonErrorResponseTransform`.

I have improved the error handling by specifically handling the `ConnectionTimeOutError`. Please note that this error can come in case of `Ollama` connection as well.

narengogi: I think it should be fine if there's multiple retries in case the host can't be reached, that is expected, checking the error by matching a string is poor code

tapansstardog: Yes @narengogi . Earlier, there were no retries in case of connection timeout. This change helps in getting expected functionality. 
Also, I have updated the code and made it specific to ConnectionTimeOutError. Please see this:

https://github.com/Portkey-AI/gateway/pull/445/files#diff-e070ec0094d3af60952ce2f32b6d12534a6d84a6a15e3f425cdf45766ef5477cR112

narengogi (approved):

tapansstardog:
@vrushankportkey @narengogi There was a conflict in the branch. I have resolved it and just pushed the changes. 
Naren, I am not sure but will you may have to approve it again to merge it into main?

VisargD (approved):

VisargD:
Continuing this in #549 to fastrack the merge. It just resolve conflicts on top of this PR.

VisargD:
The commits from this original PR are intact. Only added one more extra merge commit.

PR #460, opened 2024-07-16
Make create chat completion object strictly typed

narengogi:
**Title:** 
- Make create chat completion object type strictly typed

**Description:** (optional)
- Replicated types by looking at both the nodejs sdk of openai and the openai docs
- Gateway allows for system message to be an object which is not allowed in the latest openai sdk, gateway should deprecate this as it ads unnecessary overhead in transforming providers where both the cases have to be handled

**Testing Done**
Ran tsc --noEmit, and basic sanity. Did not remove any code

Related issue: #466

### DIFF
--- a/src/providers/anthropic/chatComplete.ts
+++ b/src/providers/anthropic/chatComplete.ts
@@ -1,10 +1,10 @@
 import { ANTHROPIC } from '../../globals';
 import {
-  Params,
-  Message,
-  ContentType,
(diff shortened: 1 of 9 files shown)


narengogi:
TO-DO: Discuss with @sk-portkey and @VisargD once on the need for these changes

narengogi:
Closing this P.R because the changes make the gateway inflexible

PR #464, opened 2024-07-17
Azure Model Name

csgulati09:
**Title:**  Azure Model Name
- Mapping azure specific headers

**Description:** (optional)
- added azure_model_name in configSchema
- handlerUtil for providerOptions
- handlerUtil mapping modelName to x-portkey-azure-model-name

**Motivation:** (optional)
- To get the custom model name

- #


src/handlers/handlerUtils.ts
```diff
@@ -198,6 +198,8 @@ export const fetchProviderOptionsFromConfig = (
       providerOptions[0].deploymentId = camelCaseConfig.deploymentId;
     if (camelCaseConfig.apiVersion)
       providerOptions[0].apiVersion = camelCaseConfig.apiVersion;
+    if (camelCaseConfig.modelName)
```
VisargD: Can we please make this `azureModelName` instead of `modelName`. For the new providers, we have started adding provider name prefix to params to make them more readable.

Issue #381, opened 2024-05-29
Support for the new `/fim/completions` route on Mistral API


vrushankportkey:
Docs - https://docs.mistral.ai/api/#operation/createFIMCompletion

keshavkrishna:
@vrushankportkey  I would like to work on this

vrushankportkey:
That's great! Assigned the issue to you @keshavkrishna!

PR #467, opened 2024-07-20
Added Support for the new /fim/completions route on Mistral AI API

keshavkrishna:
**Support new `/fim/completions `route on Mistral AI API:**

**Description:** 
- ` /fim/completions` endpoint work in the same way as the `/chat/completions` route. It handles extra  parameters such as `suffix`, `prompt`  also.
- Implemented `fimCompletionsHandler` to handle requests to the `/fim/completions `endpoint.
- Updated the API configuration to include the new `fimComplete` endpoint.
- implemented response and stream chunk transformations for the `fimComplete` function in Mistral AI.
- Modified index.ts to register the new route `v1//fim/completions` with request validation.

**Motivation:**
- To add support in portkey gateway for the new `/fim/completions` route on Mistral AI API

**Related Issues:** 
- #381


narengogi:
Thank you for the contribution @keshavkrishna will review this soon

src/providers/types.ts
```diff
@@ -153,3 +154,22 @@ export interface ImageGenerateResponse {
...
+}
+
+/**
+ * The structure of a choice in a chat completion response.
+ * @interface
+ */
+export interface FimChoice {
+  index: number;
+  message: Message;
+  finish_reason: string;
```
narengogi: Recommend making this an enum
```
enum FIM_FINISH REASON {
  stop = 'stop',
  ...
}
```

src/providers/types.ts
```diff
@@ -153,3 +154,22 @@ export interface ImageGenerateResponse {
...
+  logprobs?: object | null;
```
narengogi: Please check if the key is optional or if it can be null

narengogi (commented):
Looks good to me, will have to checkout in local and test the changes.

@VisargD do you think we should refactor the handlers to be more generic?
I think this is okay because we will not be seeing new types of inference endpoints that often, but new handlers feel redundant

VisargD:
Hey @keshavkrishna! Will you be able to take a look at the comments? And please update your branch with the latest main branch changes

keshavkrishna:
Yes I will take care of these comments by this weekend. Is that timeline okay?

src/providers/types.ts
keshavkrishna: on the mistral AI documentation it  it is written that it can be it can be array or string. what should i do. 
image

i have following two options:
1) 
```
enum FIM_FINISH_REASON {
  stop = 'stop',
}
export interface FimChoice {
  index: number;
  message: Message;
  finish_reason: FIM_FINISH_REASON;
  logprobs?: object | null;
}
```

2)

```
export interface FimChoice {
  index: number;
  message: Message;
  finish_reason: string | string[];  // Union type for string or array of strings
  logprobs?: object | null;
}

```

src/providers/types.ts
keshavkrishna: yes

VisargD:
Hey. Is it not possible to reuse the existing /chat/completions route? We can passthrough all the extra key as it is

src/providers/types.ts
keshavkrishna: done

src/providers/types.ts
keshavkrishna: removed as it wast there in mistral response object

keshavkrishna:
Requests to the existing `/chat/completions` route are redirected to the provider's base URL with the endpoint `chat/completions`. In this scenario, we need to direct requests to the provider's base URL with the endpoint `fim/completions`.

src/providers/mistral-ai/fimComplete.ts
```diff
@@ -0,0 +1,190 @@
...
+
+
+  function transformMistralFinishReasonToOpenAIFinishReason(finish_reason: MistralFinishReason | string): string {
+    switch (finish_reason) {
+        case MistralFinishReason.STOP:
+            return 'stop';
+        case MistralFinishReason.LENGTH:
+            return 'length';
+        case MistralFinishReason.MODEL_LENGTH:
+            return 'model_length';
```
narengogi: model_length is not an accepted finish reason from openai schema, you are returning the same enum as you recieve??
Please check the OpenAI documentaition to see what are the finish_reasons
image

narengogi (approved):
Approving only need to make one last change as mentioned in prev comment

VisargD:
We can accept a new header `x-portkey-mistral-fim-completion: true` and route it conditionally to the the mistral fim completions route. Adding a new router for provider specific routes is not the most feasible solution here because it will increase the number of routers in future. 

EDIT: Adding extra details
- In this function, gateway already passes providerOptions. Using it, you can check for a new param called `providerOptions.mistralFimCompletion`. If its true, you can change the endpoint to the mistral fim endpoint. 
https://github.com/Portkey-AI/gateway/blob/9724792edf7ff7954d068f00b25f90b5009a3cd0/src/providers/mistral-ai/api.ts#L8
- In the `chatComplete` config, you can add the extra fields which are required (like suffix, prompt, etc.). The response structure seems to be OpenAI compliant. So that should not be a problem

keshavkrishna:
Sure, I'll implement this

keshavkrishna:
@VisargD 

If I am not getting `mistralFimCompletion` in  `providerOptions`. following is provider option i am getting
```
{
  provider: 'mistral-ai',
  apiKey: '  SOME_API_KEY_HERE',
  overrideParams: {},
  retry: {},
  cache: {}
} 
```

so i have passed requestHeaders in `getEndpioint`, and i am checking by 
`if (requestHeaders['x-portkey-mistral-fim-completion'] === 'true') `

Is that okay??

VisargD:
> @VisargD
> 
> If I am not getting `mistralFimCompletion` in `providerOptions`. following is provider option i am getting
> 
> ```
> {
>   provider: 'mistral-ai',
>   apiKey: '  SOME_API_KEY_HERE',
>   overrideParams: {},
>   retry: {},
>   cache: {}
> } 
> ```
> 
> so i have passed requestHeaders in `getEndpioint`, and i am checking by `if (requestHeaders['x-portkey-mistral-fim-completion'] === 'true') `
> 
> Is that okay??

Hey! specific headers checks might not be required. Here is what you can do:
- In this function, add one more condition for mistral-ai provider and add the mapping for `mistralFimCompletion` in it. Check the existing conditions for other providers for reference. Once this is done, you will always get the mistralFimCompletion param in providerOptions.
https://github.com/Portkey-AI/gateway/blob/3bb3a80385258ab2b2262ac28f6e3f2f47e0e3f1/src/handlers/handlerUtils.ts#L969

Please let me know if you want an example

narengogi (commented):
Is this working? I've noticed you've not made any changes in `providers/mistral-ai/index.ts` for request and response transformers

keshavkrishna:
Yes it is. We are using chatComplete transforms as per visarg's comment.

src/providers/mistral-ai/chatComplete.ts
```diff
@@ -53,6 +53,16 @@ export const MistralAIChatCompleteConfig: ProviderConfig = {
     param: 'safe_prompt',
     default: false,
   },
+  prompt: {
+    param: 'prompt',
+    required: true,
```
VisargD: `required` should be false as it is not required for normal mistral requests.

narengogi:
> Yes it is. We are using chatComplete transforms as per visarg's comment.

for openai compliant providers, you would need to use the methods from open-ai-base provider to mention transforms
see the code for inference.net and sambanova for reference https://github.com/Portkey-AI/gateway/blob/b12ce10889497be3f3f7c9e2dd07436de0911c34/src/providers/inference-net/index.ts 
https://github.com/Portkey-AI/gateway/blob/b12ce10889497be3f3f7c9e2dd07436de0911c34/src/providers/sambanova/index.ts

src/providers/mistral-ai/chatComplete.ts
keshavkrishna: done

VisargD:
@keshavkrishna - Can you please run `npm run format` and also update with main branch? I can merge it now

keshavkrishna:
@VisargD as requested i have updated my branch and also formatted it

VisargD (approved):

Issue #116, opened 2024-01-09
Support OPEN AI spec across providers and models


raulraja:
Thank you for such an excellent library. I wanted to report some tests and discussions arising from https://discord.com/channels/1143393887742861333/1194248563199311942

The gateway currently does not support the latest format in the Open AI spec for other providers not OpenAI.

https://github.com/openai/openai-openapi/blob/master/openapi.yaml#L5421-L5429

The spec shows a message can have multiple types:
```yaml
ChatCompletionRequestMessage:
      oneOf:
        - $ref: "#/components/schemas/ChatCompletionRequestSystemMessage"
        - $ref: "#/components/schemas/ChatCompletionRequestUserMessage"
        - $ref: "#/components/schemas/ChatCompletionRequestAssistantMessage"
        - $ref: "#/components/schemas/ChatCompletionRequestToolMessage"
        - $ref: "#/components/schemas/ChatCompletionRequestFunctionMessage"
```
The content inside types like the `user` message can be strings or arrays, or message content types.
```yaml
ChatCompletionRequestUserMessage:
      type: object
      title: User message
      properties:
        content:
          description: |
            The contents of the user message.
          oneOf:
            - type: string
              description: The text contents of the message.
              title: Text content
            - type: array
              description: An array of content parts with a defined type, each can be of type `text` or `image_url` when passing in images. You can pass multiple images by adding multiple `image_url` content parts. Image input is only supported when using the `gpt-4-visual-preview` model.
              title: Array of content parts
              items:
                $ref: "#/components/schemas/ChatCompletionRequestMessageContentPart"
              minItems: 1
          x-oaiExpandable: true
        role:
          type: string
          enum: ["user"]
          description: The role of the messages author, in this case `user`.
        name:
          type: string
          description: An optional name for the participant. Provides the model information to differentiate between participants of the same role.
      required:
        - content
        - role
```
The following request:
```json
{
  "messages": [
    {
      "content": [
        {
          "text": " ...",
          "type": "text"
        }
      ],
      "role": "user"
    }
  ],
  "model": "mistralai/Mistral-7B-Instruct-v0.1",
  "max_tokens": 500,
  "temperature": 0.4,
  "user": "user"
}
```
Fails for providers `anyscale` and `together-ai` with:

Anyscale
```json
{"error":{"param":null,"code":null},"provider":"anyscale"}
```
Together AI
```json
[...]

PR #494, opened 2024-08-02
Feat: Provider Tests

narengogi:
**Title:** 
- Tests for providers to make sure they are OpenAI compliant

Initial idea for this sprung from reading the very descriptive GitHub issue raised by @raulraja https://github.com/Portkey-AI/gateway/issues/116

**Motivation:** (optional)
- Multiple provider's transformers are not OpenAI spec compliant, they fail to transform cases like user message content array
```
    {
      role: 'user',
      content: [
        {
          type: 'text',
          text: 'Who is the greatest detective',
        },
      ],
    }
```
- Already caught 6 transformers failing from 12 I tested for the two test cases.

- #116

**TO-DO:**
- [X] Push config file
- [For Later] Add flag to test only specific providers
- [For Later] GitHub actions


narengogi:
For your review @VisargD  @sk-portkey

src/tests/resources/providersConfig.ts
```diff
@@ -0,0 +1,130 @@
+import Providers from '../../providers';
+
+const providersConfig: Record<keyof typeof Providers, any> = {
```
sk-portkey: Lets strictly define type here.
ProviderTestConfig
Type ProviderTestConfig
name : keyof typeof Providers,
provider/ApiKey,
operations: LlmOperation (defined below)
skip?: boolean (if they want to skip this call)

type LlmOperation
type: 'completions' | 'chatCompletions' | 'proxy'
operation: object

src/tests/chatCompletions.test.ts
```diff
@@ -0,0 +1,50 @@
...
+} from './resources/requestTemplates';
+import transformToProviderRequest from '../services/transformToProviderRequest';
+import Providers from '../providers';
+import { Context, Hono, HonoRequest } from 'hono';
+import app from '../index';
+import providersConfig from './resources/providersConfig';
+import { CHAT_COMPLETIONS_ENDPOINT } from './resources/constants';
+import { createDefaultHeaders } from './resources/utils';
+
+for (const provider in providersConfig) {
```
sk-portkey: Instead of operation level, we can go by Provider Level. As we are aware of what each provider supports.

src/tests/resources/constants.ts
```diff
@@ -0,0 +1,2 @@
+const baseURL = 'http://localhost';
```
sk-portkey, line 1: provide an option to fetch the details from env

sk-portkey (changes requested):
move from feature specific to provider level testing

sk-portkey (dismissed):

sk-portkey:
@narengogi can you update to main. else lgtm

PR #501, opened 2024-08-12
OpenAI Audio Transcription and Translation routes support

narengogi:
**Description:** (optional)
- Add support for transcriptions and translations routes

- #488 

**TODO**
- [ ] Refactor the code

**Refactoring to do**
- [ ] Think of an alternative to hardcoding MULTIPART_FORM_DATA_ROUTES (try to check request body type)
- [ ] Theres two places at which the MULTIPART_FORM_DATA_ROUTES check is applied, make it only one
- [ ] Why are we setting content-type as multipart/formdata and removing it? let the headers be auto inferred

### DIFF
--- a/src/globals.ts
+++ b/src/globals.ts
@@ -1,4 +1,4 @@
-import { endpointStrings } from "./providers/types";
+import { endpointStrings } from './providers/types';
 
 export const POWERED_BY: string = 'portkey';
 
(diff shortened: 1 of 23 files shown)


narengogi:
changes now in https://github.com/Portkey-AI/gateway/pull/538

PR #519, opened 2024-08-17
feat: allow to accept beta for anthropic, support for prompt caching

b4s36t4:
**Title:** 
- Allow Anthropic prompt caching support in gateway.

### DIFF
--- a/src/providers/anthropic/api.ts
+++ b/src/providers/anthropic/api.ts
@@ -5,11 +5,16 @@ const AnthropicAPIConfig: ProviderAPIConfig = {
   headers: ({ providerOptions, fn }) => {
     const headers: Record<string, string> = {
       'X-API-Key': `${providerOptions.apiKey}`,
-      'anthropic-version': '2023-06-01',
     };
+
+    const betaHeader =
+      providerOptions?.['anthropicBeta'] ?? 'messages-2023-12-15';
+    const version = providerOptions?.['anthropicVersion'] ?? '2023-06-01';
+
     if (fn === 'chatComplete') {
-      headers['anthropic-beta'] = 'messages-2023-12-15';
+      headers['anthropic-beta'] = betaHeader;
     }
+    headers['anthropic-version'] = version;
     return headers;
   },
   getEndpoint: ({ fn }) => {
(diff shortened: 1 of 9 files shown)


src/types/requestBody.ts
```diff
@@ -222,6 +230,11 @@ export interface Tool {
   type: string;
   /** A description of the function. */
   function?: Function;
+
+  /**
+   * Only specific to anthropic
+   */
+  cache_control?: any;
```
VisargD: Should we use `AnthropicPromptCache` instead of `any` here?

src/providers/anthropic/api.ts
```diff
@@ -5,11 +5,15 @@ const AnthropicAPIConfig: ProviderAPIConfig = {
   headers: ({ providerOptions, fn }) => {
     const headers: Record<string, string> = {
       'X-API-Key': `${providerOptions.apiKey}`,
-      'anthropic-version': '2023-06-01',
     };
+
+    const betas = providerOptions?.['anthropicBeta'] ?? 'messages-2023-12-15';
```
VisargD: Is this a typo?

VisargD: Instead of betas, we can name it betaHeader or something similar

src/providers/anthropic/chatComplete.ts
```diff
@@ -433,7 +480,7 @@ export const AnthropicChatCompleteStreamChunkTransform: (
   chunk = chunk.trim();
 
   const parsedChunk: AnthropicChatCompleteStreamResponse = JSON.parse(chunk);
-
+  console.log(parsedChunk, 'chunk');
```
VisargD: Please remove this console log

b4s36t4: Done!

src/providers/anthropic/chatComplete.ts
```diff
@@ -118,6 +123,9 @@ export const AnthropicChatCompleteConfig: ProviderConfig = {
               const transformedMessage: Record<string, any> = {
                 role: msg.role,
                 content: [],
+                ...(msg.cache_control && {
```
VisargD: cache_control is only allowed inside the content array objects. So this mapping will not work as expected. It should be mapped inside individual objects

VisargD (approved):

PR #533, opened 2024-08-21
[Provider]: Deepbricks, SiliconFlow

sternelee:
support [deepbricks](https://deepbricks.ai/), [siliconflow](https://siliconflow.cn/)
They provide a more convenient and comprehensive LLMS service

### DIFF
--- a/src/providers/siliconflow/imageGenerate.ts
+++ b/src/providers/siliconflow/imageGenerate.ts
@@ -1,0 +1,47 @@
+import { SILICONFLOW } from '../../globals';
+import { ErrorResponse, ImageGenerateResponse, ProviderConfig } from '../types';
+import { SiliconFlowErrorResponseTransform } from './chatComplete';
+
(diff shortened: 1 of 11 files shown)


src/providers/siliconflow/imageGenerate.ts
```diff
@@ -0,0 +1,47 @@
+import { SILICONFLOW } from '../../globals';
+import { ErrorResponse, ImageGenerateResponse, ProviderConfig } from '../types';
+import { SiliconFlowErrorResponseTransform } from './chatComplete';
+
+export const SiliconFlowImageGenerateConfig: ProviderConfig = {
+  prompt: {
+    param: 'prompt',
+    required: true,
+  },
+  image_size: {
```
VisargD: We take the input in OpenAI format. So this should take size as input and map it to image_size

VisargD (dismissed):

Issue #537, opened 2024-08-24
[Feature] Support Audio routes for Azure OpenAI


narengogi:
### What Would You Like to See with the Gateway?

add new model support

PR #538, opened 2024-08-24
Azure OpenAI  Audio routes

narengogi:
- https://github.com/Portkey-AI/gateway/issues/537

### DIFF
--- a/src/handlers/handlerUtils.ts
+++ b/src/handlers/handlerUtils.ts
@@ -14,7 +14,7 @@ import {
 } from '../globals';
 import Providers from '../providers';
 import { ProviderAPIConfig, endpointStrings } from '../providers/types';
-import transformToProviderRequestJSON from '../services/transformToProviderRequest';
+import transformToProviderRequest from '../services/transformToProviderRequest';
 import {
   Config,
   Options,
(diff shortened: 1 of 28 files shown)


src/handlers/handlerUtils.ts
```diff
@@ -513,7 +540,11 @@ export async function tryPost(
     requestHeaders
   );
 
-  fetchOptions.body = JSON.stringify(transformedRequestBody);
+  fetchOptions.body = MULTIPART_FORM_DATA_ENDPOINTS.includes(fn)
+    ? (transformedRequestBody as FormData)
+    : JSON.stringify(transformedRequestBody);
+
+  console.log('fetchOptions', fetchOptions);
```
VisargD: Please remove this log

src/handlers/handlerUtils.ts
```diff
@@ -67,9 +69,13 @@ export function constructRequest(
     method,
     headers,
   };
+  const contentType = headers['content-type'];
+  const isGetMethod = method === 'GET';
```
VisargD, line 76: The code will never reach here if it is a GET method. So we can remove it

VisargD (approved):

PR #565, opened 2024-08-30
refactor code for minimal changes required for new open-ai compatible…

b4s36t4:
Simplifying the creation of new openai compatible providers.

**API**
* `chatCompletetionParams` - Extendable function to support exclusion, default-values and extra config based on the requirement for `chat` functionality of provider.
* `completionParams` - Extendable function to support exclusion, default-values and extra config based on the requirement for `chat` functionality of provider.
* `embedParams` - Extendable function to support exclusion, default-values and extra config based on the requirement for `embed` functionality of provider.

**Response Handling**
`responseTransformers` function accepts two parameter
* `provider` as string
* `options` an object - `{chatComplete: Function | boolean, complete: Function | boolean, embed: Function | boolean}`

We can enable `openai` existing response transformers directly by passing `true` for each function or we can write a function that accepts `response` and `boolean` indicating weather the response is error response or not.

Extended function should return the response which is inclusive of `openai` response types (embed, complete and chatComplete).

### DIFF
--- a/src/globals.ts
+++ b/src/globals.ts
@@ -66,7 +66,7 @@ export const VOYAGE: string = 'voyage';
 export const GITHUB: string = 'github';
 export const DEEPBRICKS: string = 'deepbricks';
 export const SILICONFLOW: string = 'siliconflow';
-export const CREEBRAS: string = 'cerebras';
+export const CEREBRAS: string = 'cerebras';
 
 export const VALID_PROVIDERS = [
   ANTHROPIC,
@@ -107,7 +107,7 @@ export const VALID_PROVIDERS = [
   DEEPBRICKS,
   SILICONFLOW,
   HUGGING_FACE,
-  CREEBRAS,
+  CEREBRAS,
 ];
 
 export const CONTENT_TYPES = {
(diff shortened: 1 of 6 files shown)


src/globals.ts
```diff
@@ -66,6 +66,7 @@ export const VOYAGE: string = 'voyage';
 export const GITHUB: string = 'github';
 export const DEEPBRICKS: string = 'deepbricks';
 export const SILICONFLOW: string = 'siliconflow';
+export const CREEBRAS: string = 'cerebras';
```
VisargD: Can we fix this typo :)

src/providers/open-ai-base/index.ts
```diff
@@ -0,0 +1,378 @@
...
+  }
+
+  if (options.complete) {
+    transformers.complete = CompleteResponseTransformer<U>(provider);
+  }
+
+  if (options.chatComplete) {
+    transformers.chatComplete = ChatCompleteResponseTransformer<V>(
+      provider,
+      typeof options.chatComplete === 'function'
```
VisargD, line 378: I think we can have the `function` type check for all other transformers also, right?

b4s36t4: Yep, they're not extended. Forget to extend them would do the needful.

VisargD:
@b4s36t4 - Is this ready to review?

VisargD (approved):

Issue #457, opened 2024-07-11
Anthropic only uses Last System Message


jpulec:
### What Happened?

The way Portkey is set up, when sending multiple system messages to anthropic, only the last message is sent.

### What Should Have Happened?

I'd propose that a better solution, since anthropic only allows a single system message to be provided, would be to concatenate all the provided system messages together, instead of only sending the last one.

If this seems reasonable, I can write the code for it.

### Relevant Code Snippet

Note how system message is just wholesale overwritten in every iteration of `param,messages`:

https://github.com/Portkey-AI/gateway/blob/fe0899bbf1a0f65f8f751aca8fca1b0a4409acb4/src/providers/anthropic/chatComplete.ts#L174-L186

### Your Twitter/LinkedIn

@jamespulec

vrushankportkey:
Thank you @jpulec - this seems interesting and useful. Tagging @VisargD / @roh26it who can give the final go.

roh26it:
Hi @jpulec - Sounds like the right approach, can you raise a PR for this. Would be happy to merge this.

PR #575, opened 2024-09-05
feat: combine system messages for anthropic

jpulec:
**Title:** 
Combine all system prompts into one prompt for anthropic.

**Motivation:**
- Most other providers support multiple system prompts, so this makes anthropic more consistent with other providers.

**Related Issues:**
- #457


narengogi (approved):
Looks good, nice minimal fix!

VisargD (approved):

Issue #597, opened 2024-09-11
[Feature] Add Provider upstage.ai


b4s36t4:
### What Would You Like to See with the Gateway?

Support upstage.ai as a inference provider. [Docs](https://developers.upstage.ai/docs/getting-started/quick-start)

PR #608, opened 2024-09-13
Upstage.ai

shivam-pareek:
**Title:** 
- Added upstage.ai as a provider 

**Description:** (optional)
- chat completion endpoint
- embeddings endpoint

- #597


src/providers/index.ts
```diff
@@ -42,6 +42,7 @@ import HuggingfaceConfig from './huggingface';
 import { cerebrasProviderAPIConfig } from './cerebras';
 import { InferenceNetProviderConfigs } from './inference-net';
 import SambaNovaConfig from './sambanova';
+import { UpstageProviderConfig } from './upstage';
```
narengogi: Please rename this to UpstageConfig to keep with the convention we've been following above (I can see InferenceNet and cerebras are also named this way, but those have to be changed as well)

narengogi (approved):
LGTM, commented on variable renaming required

VisargD (approved):

PR #615, opened 2024-09-17
Feat: Stable diffusion v3

narengogi:
**Description:**
- Added support for Stable diffusion v2Beta route with support for 3 text-to-image models. Core, Ultra and 3Large
- Same support in Bedrock also
- Added a new parameter in provider config `transformToFormData` using this to transform to form data
- Added the new transformer for form data also

- https://github.com/Portkey-AI/gateway/issues/616

**Testing Done:**
- [X] Tested to make sure SDXL-1 is working
- [X] verified error and success response transforms
- [X] verified by making requests to /audio/translations and /transcriptions

### DIFF
--- a/src/globals.ts
+++ b/src/globals.ts
@@ -135,10 +135,6 @@ export const CONTENT_TYPES = {
   GENERIC_IMAGE_PATTERN: 'image/',
 };
 
-export const ACCEPT_TYPES = {
-  APPLICATION_JSON: 'application/json',
-};
-
 export const MULTIPART_FORM_DATA_ENDPOINTS: endpointStrings[] = [
   'createTranscription',
   'createTranslation',
--- a/src/globals.ts
+++ b/src/globals.ts
@@ -13,6 +13,7 @@ export const HEADER_KEYS: Record<string, string> = {
   CUSTOM_HOST: `x-${POWERED_BY}-custom-host`,
   REQUEST_TIMEOUT: `x-${POWERED_BY}-request-timeout`,
   STRICT_OPEN_AI_COMPLIANCE: `x-${POWERED_BY}-strict-open-ai-compliance`,
+  CONTENT_TYPE: `content-type`,
 };
 
 export const RESPONSE_HEADER_KEYS: Record<string, string> = {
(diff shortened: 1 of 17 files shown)


src/types/requestBody.ts
```diff
@@ -136,6 +143,8 @@ export interface Targets {
   index?: number;
   cache?: CacheSettings | string;
   targets?: Targets[];
+  /** This is used to determine if the request should be transformed to formData Example: Stability V2 */
+  transFormToFormData?: boolean;
```
VisargD: Can we rename this to `transformToFormData` (lower case f in transform)

narengogi: done

src/globals.ts
```diff
@@ -126,6 +127,10 @@ export const CONTENT_TYPES = {
   GENERIC_IMAGE_PATTERN: 'image/',
 };
 
+export const ACCEPT_TYPES = {
```
VisargD: Can we reuse the `CONTENT_TYPES` global instead of creating a new one?

src/providers/bedrock/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const BEDROCK_STABILITY_V2_MODELS = [
```
VisargD: Considering that v1 models are now deprecated, should we maintain a list of those models instead of maintaining v2 models? Because v2 models might keep increasing as they are the active version.

VisargD: Or a better solution as per me would be to maintain naming patterns. For example, anything that contains `sd3` in it is the new model

VisargD:
@narengogi - Can you please also check if fireworks imageGenerate needs an update for this?

src/providers/bedrock/constants.ts
narengogi: will check fro v1 models since they are limited

src/handlers/imageGenerationsHandler.ts
```diff
@@ -18,6 +19,12 @@ export async function imageGenerationsHandler(c: Context): Promise<Response> {
     let request = await c.req.json();
     let requestHeaders = Object.fromEntries(c.req.raw.headers);
     const camelCaseConfig = constructConfigFromRequestHeaders(requestHeaders);
+    if (
```
VisargD: Can we move this to provider configs instead of top-level handlers?

VisargD (approved):

PR #618, opened 2024-09-17
Add missing logprob mapping for Azure OpenAI provider

rootAvish:
**Title:** Add missing logprob mapping for Azure OpenAI provider

**Description:** The Portkey gateway doesn't return any logprobs when these are set to True and 5 respectively, calling Azure directly works. I guess the reason is this mapping was missing. Kindly merge this so we can get logprobs when calling Azure via Portkey.

**Motivation:** We use the SaaS version of your gateway @ Haptik, and I need these scores.


VisargD (approved):

rootAvish:
Thanks for the quick turnaround on the review @VisargD. Do I need to do anything else here? If not, can we please merge this?

PR #647, opened 2024-10-01
feat: pass provider by model name

sternelee:
By passing the provider through the model parameter without needing to change the request header, it can be more conveniently used in various AI code editors and other AI services.


vrushankportkey:
@VisargD / @roh26it thoughts on this PR?

roh26it:
It's definitely an interesting way to think about it. 

Curious why @sternelee closed it

roh26it:
@VisargD any thoughts on this?

VisargD:
Hi, closing this PR because we are planning to add support of provider configs and alias in Open source gateway. With that change, we will also allow using `<provider-alias>/<model-name>` in the model field. We will add more details on it once the changes are finalized on our end. Thanks!

Issue #650, opened 2024-10-03
[Feature] Add Lambda Chat as a new provider


b4s36t4:
### What Would You Like to See with the Gateway?

Integrate Lambda Chat completions API as a provider to gateway.

[Docs](https://docs.lambdalabs.com/on-demand-cloud/using-the-lambda-chat-completions-api#creating-completions)

narengogi:
closing as stale

PR #651, opened 2024-10-03
feat: lambda provider

b4s36t4:
**Title:** 
- Adds lambda chat as new provider to gateway.

- #650

### DIFF
--- a/src/providers/lambda/api.ts
+++ b/src/providers/lambda/api.ts
@@ -5,7 +5,7 @@ export const LambdaAPIConfig: ProviderAPIConfig = {
   headers({ providerOptions }) {
     const { apiKey } = providerOptions;
     const headers = {
-      Authorization: apiKey,
+      Authorization: `Bearer ${apiKey}`,
     };
     return headers;
   },
--- a/src/providers/lambda/api.ts
+++ b/src/providers/lambda/api.ts
@@ -1,0 +1,24 @@
+import { ProviderAPIConfig } from '../types';
+
+export const LambdaAPIConfig: ProviderAPIConfig = {
+  getBaseURL: () => `https://api.lambdalabs.com/v1`,
+  headers({ providerOptions }) {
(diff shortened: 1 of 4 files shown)


src/providers/lambda/api.ts
```diff
@@ -0,0 +1,24 @@
+import { ProviderAPIConfig } from '../types';
+
+export const LambdaAPIConfig: ProviderAPIConfig = {
+  getBaseURL: () => `https://api.lambdalabs.com/v1`,
+  headers({ providerOptions }) {
+    const { apiKey } = providerOptions;
+    const headers = {
+      Authorization: apiKey,
```
VisargD: As per their docs examples, `Bearer` is required in the `Authorization` header. Can we please add it here as well?

VisargD: https://docs.lambdalabs.com/on-demand-cloud/using-the-lambda-chat-completions-api#creating-chat-completions

src/providers/lambda/index.ts
```diff
@@ -0,0 +1,18 @@
...
+  chatCompleteParams,
+  completeParams,
+  responseTransformers,
+} from '../open-ai-base';
+import { ProviderConfigs } from '../types';
+import { LambdaAPIConfig } from './api';
+
+export const LambdaProviderConfig: ProviderConfigs = {
+  chatComplete: chatCompleteParams([], { model: 'Liquid-AI-40B' }),
+  complete: completeParams([], { model: 'Liquid-AI-40B' }),
```
VisargD, line 12: Is this the correct model ID? Their docs does not seem to mention this exact id anywhere. So I am not sure.

b4s36t4: I took it from their playground. They do support the `liquid` models as well.

VisargD: Got it

VisargD (approved):

PR #655, opened 2024-10-03
Adding metadata to webhook guardrails

roh26it:
# Add metadata to HookSpanContext and webhook plugin

## Changes
This pull request introduces the following changes to the AI gateway repo:

1. Added metadata to the HookSpanContext, which is now available as `plugincontext` in all plugins.
2. Updated the webhook plugin to accept a new `sendMetadata` option, allowing metadata to be sent along with the webhook information.

## Purpose
These changes enhance the flexibility and information sharing capabilities of our plugin system, particularly for the webhook plugin.


plugins/default/webhook.ts
```diff
@@ -20,7 +20,11 @@ export const handler: PluginHandler = async (
   let data = null;
   try {
     let url = parameters.webhookURL;
-    const headers = parseHeaders(parameters.headers);
+
+    let headers: Record<string, string> = parameters?.headers
```
VisargD: Hey. Looks like this got changed while resolving some merge conflicts? We can keep the existing `parseHeaders` function call because it handles both: object and stringified headers.

roh26it: Damn, sorry about this! Reverted it back

VisargD (approved):

PR #656, opened 2024-10-03
Adding explanations to default guardrails

roh26it:
**Adding explanations to the data property of default guardrail plugins** 
We're adding a `data.explanation` property to default guardrails to make it easier to identify why they succeeded or failed.


VisargD:
Changing this to draft. Please update the PR once its ready

roh26it:
@VisargD all tests for default plugins are passing. You'll need to check the portkey plugins with credentials. I could not run those tests

package.json
```diff
@@ -46,6 +46,8 @@
     "@portkey-ai/mustache": "^2.1.2",
     "@smithy/signature-v4": "^2.1.1",
     "@types/mustache": "^4.2.5",
+    "ajv": "^8.17.1",
```
VisargD: From what I remember, `ajv` does not support edge runtimes. Let me confirm once.

VisargD: Also, any reason to replace zod with ajv?

roh26it: Replaced it with worker compatible json schema validator now

VisargD (approved):

PR #677, opened 2024-10-11
Feat: portley llm guardrails update

VisargD:
**Title:** 
- Allow gateway deployments to use Portkey implemented LLM-based guardrails like pii, language detection, etc.

**Description:** (optional)
- Pass env object to handler functions so that it can be used to pick up variables.
- Add is_enabled flag on check level to allow disabling individual checks without removing it from the hooks object.

**Motivation:** (optional)
- Enable portkey llm guardrails

- #issue-number

### DIFF
--- a/plugins/portkey/gibberish.ts
+++ b/plugins/portkey/gibberish.ts
@@ -10,7 +10,8 @@ import { PORTKEY_ENDPOINTS, fetchPortkey } from './globals';
 export const handler: PluginHandler = async (
   context: PluginContext,
   parameters: PluginParameters,
-  eventType: HookEventType
+  eventType: HookEventType,
+  options
 ) => {
   let error = null;
   let verdict = false;
@@ -22,6 +23,7 @@ export const handler: PluginHandler = async (
 
     // Check if the text is gibberish
     const response: any = await fetchPortkey(
+      options.env,
       PORTKEY_ENDPOINTS.GIBBERISH,
       parameters.credentials,
       { input: text }
--- a/src/middlewares/hooks/index.ts
+++ b/src/middlewares/hooks/index.ts
@@ -8,6 +8,7 @@ import {
   HookOnFailObject,
   HookOnSuccessObject,
   HookResult,
+  HandlerOptions,
 } from './types';
 import { plugins } from '../../../plugins';
 import { Context } from 'hono';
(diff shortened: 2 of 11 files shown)


plugins/portkey/gibberish.ts
```diff
@@ -10,7 +10,8 @@ import { PORTKEY_ENDPOINTS, fetchPortkey } from './globals';
 export const handler: PluginHandler = async (
   context: PluginContext,
   parameters: PluginParameters,
-  eventType: HookEventType
+  eventType: HookEventType,
+  options
```
b4s36t4, line 14: Can't we add `ts` support for this as well?!

src/middlewares/hooks/index.ts
```diff
@@ -296,9 +303,16 @@ export class HooksManager {
 
     if (hook.type === 'guardrail' && hook.checks) {
       const checkResults = await Promise.all(
-        hook.checks.map((check: Check) =>
-          this.executeFunction(span.getContext(), check, hook.eventType)
-        )
+        hook.checks
+          .filter((check: Check) => check.is_enabled !== false)
```
b4s36t4, line 307: `is_enabled` is an optional field which maybe `undefined` and that can cause this condition true. Can you check?!

b4s36t4 (changes requested):

plugins/portkey/gibberish.ts
VisargD: The handler already has `PluginHandler` type. So it will automatically infer options type from PluginHandler. Felt redundant to add it explicitly. Did not want to remove it from existing args in this PR.

VisargD: Screenshot 2024-10-11 at 19 42 59

src/middlewares/hooks/index.ts
VisargD: Hey! The default behaviour is to assume all checks to be is_enabled true. So we are specifically checking for false to verify if it needs to be disabled. Do you think this will cause any issues?

b4s36t4: As per the ts interface for `Check` `is_enabled` can be `undefined`. Now `undefined !== false` is resulting `true`.

If we're sure that `is_enabled` is always available,  we can remove optional specifier in type def.

VisargD: is_enabled is not a mandatory field. If its not set, we assume the default behaviour which is true. Until its explicitly set as false, we will consider it as valid

PR #678, opened 2024-10-14
Feature: Bedrock converse integration

narengogi:
**Title:** 
- This replaces the need to have a tokenizer for open source models hosted on AWS bedrock, converse is bedrock's unified gateway like implementation

Limitations:
- Few legacy models are not supported in the converse endpoint (https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference-supported-models-features.html)

**Testing Done:**
- [X] Verified stream chunk handing of older routes
- [X] Tool calling and Tool Responses in chat messages, single and multi turn
- [X] Chat Completions
- [X] Tested with different models
- [X] Tested prompting with images

- #625

### DIFF
--- a/src/providers/bedrock/chatComplete.ts
+++ b/src/providers/bedrock/chatComplete.ts
@@ -469,7 +469,7 @@ export const BedrockConverseAnthropicChatCompleteConfig: ProviderConfig = {
       transformAnthropicAdditionalModelRequestFields(params),
   },
   user: {
-    param: 'user',
+    param: 'additionalModelRequestFields',
     transform: (params: BedrockConverseAnthropicChatCompletionsParams) =>
       transformAnthropicAdditionalModelRequestFields(params),
   },
--- a/src/providers/bedrock/chatComplete.ts
+++ b/src/providers/bedrock/chatComplete.ts
@@ -455,22 +455,22 @@ export const BedrockConverseAnthropicChatCompleteConfig: ProviderConfig = {
   ...BedrockConverseChatCompleteConfig,
   additionalModelRequestFields: {
     param: 'additionalModelRequestFields',
-    transform: (params: BedrockChatCompletionsParams) =>
+    transform: (params: BedrockConverseAnthropicChatCompletionsParams) =>
       transformAnthropicAdditionalModelRequestFields(params),
   },
   top_k: {
     param: 'additionalModelRequestFields',
-    transform: (params: BedrockChatCompletionsParams) =>
+    transform: (params: BedrockConverseAnthropicChatCompletionsParams) =>
       transformAnthropicAdditionalModelRequestFields(params),
   },
   anthropic_version: {
     param: 'additionalModelRequestFields',
-    transform: (params: BedrockChatCompletionsParams) =>
+    transform: (params: BedrockConverseAnthropicChatCompletionsParams) =>
       transformAnthropicAdditionalModelRequestFields(params),
   },
   user: {
     param: 'user',
-    transform: (params: BedrockChatCompletionsParams) =>
+    transform: (params: BedrockConverseAnthropicChatCompletionsParams) =>
       transformAnthropicAdditionalModelRequestFields(params),
   },
 };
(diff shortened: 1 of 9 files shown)


src/providers/bedrock/index.ts
```diff
@@ -56,108 +40,89 @@ import {
 const BedrockConfig: ProviderConfigs = {
   api: BedrockAPIConfig,
   getConfig: (params: Params) => {
-    if (!params.model) {
```
VisargD: Can you please check why this code was removed?

src/providers/bedrock/chatComplete.ts
```diff
@@ -830,176 +245,103 @@ export const BedrockAI21ChatCompleteResponseTransform: (
...
         {
+          index: 0,
           message: {
             role: 'assistant',
-            content,
-            tool_calls: toolCalls.length ? toolCalls : undefined,
+            content: response.output.message.content
+              .filter((content) => content.text)
+              .reduce((acc, content) => acc + content.text + '\n', ''),
+            tool_calls: response.output.message.content
```
VisargD: Due to this logic, the response will always have tool_calls as empty array when the response does not have any tool calls. Can we make it optional?

src/providers/bedrock/chatComplete.ts
```diff
@@ -9,613 +9,218 @@ import {
...
+              return acc.concat(...getMessageTextContentArray(msg));
+            return acc;
+          },
+          []
+        );
+        if (!systemMessages.length) return;
+        return systemMessages;
+      },
+    },
+    {
```
VisargD: These params should be independently mapped. Currently additionalModelRequestFields and inferenceConfig are added as a part of messages transform logic

src/providers/bedrock/chatComplete.ts
VisargD, line 1: Not sure if I am missing something but I don't see any handling for multimodal prompting

src/providers/bedrock/index.ts
```diff
@@ -56,108 +40,89 @@ import {
...
-      throw new GatewayError('Bedrock model not found');
-    }
-
-    // To remove the region in case its a cross-region inference profile ID
-    // https://docs.aws.amazon.com/bedrock/latest/userguide/cross-region-inference-support.html
-    const providerModel = params.model.replace(/^(us\.|eu\.)/, '');
```
VisargD: This is important to support cross region model IDs. Please add the whole code block back.

src/providers/bedrock/chatComplete.ts
```diff
@@ -9,613 +9,218 @@ import {
...
+        return {
+          maxTokens: params.max_tokens || params.max_completion_tokens,
+          stopSequences:
+            typeof params.stop === 'string' ? [params.stop] : params.stop,
+          temperature: params.temperature,
+          topP: params.top_p,
+        };
+      },
+    },
+    {
```
VisargD: Previously, we had individual mapping for some provider specific params like anthropic_version. With converse API, these needs to go into  `additionalModelRequestFields`. But from user perspective, we should keep it backward compatible. So they can still send anthropic_version and gateway should map it to additional fields

src/providers/bedrock/index.ts
narengogi: Adding it back, (lost some changes in a messy rebase) I'll check at other places also

src/providers/bedrock/chatComplete.ts
narengogi: I think this is the correct approach
InferenceConfig has the following fields
```
          maxTokens: params.max_tokens || params.max_completion_tokens,
          stopSequences:
            typeof params.stop === 'string' ? [params.stop] : params.stop,
          temperature: params.temperature,
          topP: params.top_p,
        };
```
How would you recommend mapping this? If I map it under max_tokens, than would mean that max_tokens has to be present for me to map top_p.

VisargD: All the params should have their individual mapping so that its readable on top-level. All of these can be separate mapping which transforms to `InferenceConfig.*`. You can take a reference from vertex mapping.

src/providers/bedrock/chatComplete.ts
narengogi: please check line: 70 in `bedrock/chatCompletions.ts`

src/providers/bedrock/index.ts
```diff
@@ -59,105 +49,109 @@ const BedrockConfig: ProviderConfigs = {
     if (!params.model) {
       throw new GatewayError('Bedrock model not found');
     }
-
```
VisargD: Please add this back. Its better for readability.

src/providers/bedrock/chatComplete.ts
```diff
@@ -13,610 +13,270 @@ import {
...
+    message.content.forEach((item) => {
+      if (item.type === 'text') {
+        out.push({
+          text: item.text || '',
+        });
+      } else if (item.type === 'image_url' && item.image_url) {
+        const imageParts = item.image_url.url.split(';');
+        const imageFormat = imageParts[0].split('/')[1];
+        out.push({
+          image: {
```
VisargD: Converse supports `document` and `image` both. Based on the the mimetype, can we decide what field should be used?

https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference-call.html

src/providers/bedrock/chatComplete.ts
```diff
@@ -13,610 +13,284 @@ import {
...
-        });
-        prompt += 'Assistant:';
-      }
-      return prompt;
-    },
+  additionalModelResponseFieldPaths: {
+    param: 'additionalModelResponseFieldPaths',
+    required: false,
   },
   max_tokens: {
```
VisargD, line 226: Can you please also add mapping for max_completion_tokens

src/providers/bedrock/utils.ts
```diff
@@ -45,99 +40,55 @@ export const generateAWSHeaders = async (
...
+  params: BedrockChatCompletionsParams
+) => {
+  const additionalModelRequestFields: Record<string, any> =
+    params.additionalModelRequestFields || {};
+  if (params['top_k']) {
+    additionalModelRequestFields['topK'] = params['top_k'];
+  }
+  // Backward compatibility
+  if (params['anthropic_version']) {
+    additionalModelRequestFields['anthropic_version'] =
```
VisargD, line 88: As per Bedrock Converse docs, additionalModelRequestFields just needs to have path of the param and not the actual param value:
```
You can specify the paths for additional model parameters in the additionalModelResponseFieldPaths field, as shown in the following example.


[ "/stop_sequence" ]
The API returns the additional fields that you request in the additionalModelResponseFields field.
```

VisargD: Please ignore. Confused `additionalModelRequestFields` with ResponseFields

VisargD (approved):

Issue #674, opened 2024-10-09
[Feature] Valid Numerical Range


ashirwad19:
### What Would You Like to See with the Gateway?

Plugin Name: Valid Numerical Range
Plugin Description: Validates whether the generated number is in the acceptable range.

Here is the [Contributing Guide for Plugin](https://github.com/Portkey-AI/gateway/blob/main/plugins/Contributing.md)

nikhil-kumar94:
@ashirwad19  Please assign me this issue.

narengogi:
Hey @nikhil-kumar94, you can start working on any issue you would want to contribute on and add a comment in the corresponding issue that you are working on it

nikhil-kumar94:
@ashirwad19 can you please review this PR 
PR :- https://github.com/Portkey-AI/gateway/pull/679

PR #679, opened 2024-10-14
added valid number range

nikhil-kumar94:
**Title:** 
- Valid Number range plugin

**Related Issues:**
- #674

### DIFF
--- a/plugins/default/validNumberRange.ts
+++ b/plugins/default/validNumberRange.ts
@@ -1,0 +1,43 @@
+import {
+  HookEventType,
+  PluginContext,
+  PluginHandler,
+  PluginParameters,
+} from '../types';
+import { getText } from '../utils';
+
+function validNumberRange(
+  target: Number,
+  lowerLimit: Number,
+  upperLimit: Number
+): boolean {
+  try {
+    return target >= lowerLimit && target <= upperLimit;
+  } catch (error) {
+    return false;
+  }
+}
+
+export const handler: PluginHandler = async (
+  context: PluginContext,
+  parameters: PluginParameters,
+  eventType: HookEventType
+) => {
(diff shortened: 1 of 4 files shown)


plugins/default/validNumberRange.ts
```diff
@@ -0,0 +1,43 @@
...
+  PluginContext,
+  PluginHandler,
+  PluginParameters,
+} from '../types';
+import { getText } from '../utils';
+
+function validNumberRange(
+  target: Number,
+  lowerLimit: Number,
+  upperLimit: Number
```
b4s36t4, line 12: `Number` isn't a type. use `number` instead.

plugins/default/validNumberRange.ts
```diff
@@ -0,0 +1,43 @@
...
+): boolean {
+  try {
+    return target >= lowerLimit && target <= upperLimit;
+  } catch (error) {
+    return false;
+  }
+}
```
b4s36t4, line 19: I don't think this logic is what the expectation is. Ideally plugins are used as part of generation so we must be able to use this rule with generated content. This rule must check weather the generated content (number) is within the range.

use `getText` function to retrieve the message content and make sure the content is number (or possibly use regex to extract number from the message content) and do validation with the extracted number.

Check existing plugins here: https://github.com/Portkey-AI/gateway/tree/main/plugins/default

b4s36t4 (changes requested):

VisargD:
Hey @nikhil-kumar94 and @b4s36t4 , is there any blocker for this PR? Please let me know if you have any doubts.

VisargD:
Closing this PR due to inactivity. Please feel free to reopen it whenever required

PR #692, opened 2024-10-21
Add a help page & logging when gateway starts

roh26it:
**Help Page and Logs** 
- This PR introduces a webpage that opens up when the gateway is started through node.js
- The webpage contains help information on using the gateway
- The logs tab provides realtime running logs for requests made through the gateway

**How to try it out**
- Pull this branch and run the gateway normally through the `npm run dev:node` command.

**Motivation:** 
- Besides helping new users to consume gateway features fast, it also helps identify bugs faster through the logging section.


roh26it:
Tested docker, npx deployments

src/middlewares/log/index.ts
```diff
@@ -0,0 +1,87 @@
...
+    const ms = Date.now() - start;
+    if (!c.req.url.includes('/v1/')) return;
+
+    const requestOptionsArray = c.get('requestOptions');
+    if (requestOptionsArray[0].requestParams.stream) {
+      requestOptionsArray[0].response = {
+        message: 'The response was a stream.',
+      };
+    } else {
+      const response = await c.res.clone().json();
```
VisargD: I think this will add latency to the actual response because the client will not receive the response until this whole code gets executed. Worth checking before merging this.

roh26it: Yes, the clone and broadcastLog will be executed along with the request. I'm hoping this only happens when someone runs this locally. Is there a way to move it to async?

VisargD: > I'm hoping this only happens when someone runs this locally

There is no check for this part in code. So it will get executed for all the requests.

roh26it: The logger middleware is attached only in start-server.ts and when there's no headless mode Param passed.

VisargD: For `getRuntimeKey() === "workerd"`, we can use `c.executionCtx.waitUntil()`

And for `getRuntimeKey() === "node"`, we can use `.then().catch()`. This one needs to be verified once.

VisargD:
@roh26it - Should we also put a env or some flag check to enable the help page and logging? For example, `production` env might not need the UI.

roh26it:
Aah, yes we could do that. Let me add a production check to env as well.

roh26it:
@VisargD made the requested changes

VisargD (approved):

PR #695, opened 2024-10-22
[New plugin] Simple plugin to allow whitelisting models as a guardrail

elentaure:
**Title:** 
- This PR adds a simple plugin to allow defining a list of allowed models as a guardrail. Please, review and let me know any feedback


narengogi (approved):
LGTM

PR #719, opened 2024-10-30
Forward params not in the provider config keys

narengogi:
…
**Title:** 
- transformToProviderConfig has been looping over the provider config and setting keys, Flipped it to loop over request params

**Motivation:** (optional)
- To forward any params not specified in the config without transformation

- #issue-number

**Testing done**
- [X] Tested when transform object is a recursive array
- [X] Tested nested properties
- [X] Tested value as 'default' 
- [X] Tested when required and default is available
- [X] Tested remaining routes for sanity

**Note**
I've not added support for min and max

### DIFF
--- a/src/services/transformToProviderRequest.ts
+++ b/src/services/transformToProviderRequest.ts
@@ -1,7 +1,7 @@
 import { GatewayError } from '../errors/GatewayError';
 import { MULTIPART_FORM_DATA_ENDPOINTS } from '../globals';
 import ProviderConfigs from '../providers';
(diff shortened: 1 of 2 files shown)


narengogi:
Closing this after discussing with @VisargD @roh26it @ayush-portkey 

**summary:** 
1. These changes are not backward compatible if clients have already been sending additional parameters and they've been getting ignored till now
2. This will restrict dynamic-routing/fallbacks between providers requiring users to override params to null or undefined instead of forwarding them (@VisargD)

Issue #512, opened 2024-08-16
[Feature] Endpoint Is Reachable Plugin


siddharthsambharia-portkey:
### What Would You Like to See with the Gateway?

Plugin Name: Endpoint Is Reachable
Plugin Description: Checks if an endpoint can be reached by making a request to it.

Here is the [Contributing Guide for Plugin](https://github.com/Portkey-AI/gateway/blob/main/plugins/Contributing.md)

s-vamshi:
@VisargD I would like to work on this issue!

vrushankportkey:
Hi @s-vamshi please go ahead and start working on this!

5war00p:
Can I take a stab at this, as I see no progress made from 2 weeks.

vrushankportkey:
Hey! Yes, please, go ahead!

5war00p:
@vrushankportkey PR is ready for review.

vrushankportkey:
Thank you!

PR #720, opened 2024-10-31
feat: endpoint reachability plugin

5war00p:
**Endpoint Reachability Plugin** 
- Brief Description of Changes
Added new plugin that checks and validates whether an endpoint (url) is reachable or unreachable.

**Motivation:** (optional)
Its a preventive check to not do unnecessary actions without a proper input

- #512

### DIFF
--- a/plugins/default/endpointReachability.ts
+++ b/plugins/default/endpointReachability.ts
@@ -1,0 +1,34 @@
+import { PluginHandler } from '../types';
+import { getText } from '../utils';
+
+const endpointReachability = async (url: string): Promise<boolean> => {
+  if (!url) return false
+  try {
+    // Making a HEAD request as its lightweight (gets headers and doesn't wait for response)
+    const response = await fetch(url, { method: 'HEAD' });
+    return response.ok;
+  } catch (error) {
+    // On all other cases like 403, 404, 500
+    return false;
+  }
+};
+
+export const handler: PluginHandler = async (
+  context,
+  _parameters,
+  eventType,
+  _options
+) => {
+  let error = null;
+  let verdict = false;
+  let data = null;
+
+  try {
+    let text = getText(context, eventType);
+    verdict = await endpointReachability(text);
+  } catch (e) {
+    error = e as Error;
+  }
+
+  return { error, verdict, data };
+};
--- a/plugins/default/endpointReachability.ts
+++ b/plugins/default/endpointReachability.ts
@@ -2,7 +2,7 @@ import { PluginHandler } from '../types';
 import { getText } from '../utils';
 
 const endpointReachability = async (url: string): Promise<boolean> => {
-  if (!url) return false
+  if (!url) return false;
   try {
     // Making a HEAD request as its lightweight (gets headers and doesn't wait for response)
     const response = await fetch(url, { method: 'HEAD' });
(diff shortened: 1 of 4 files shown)


roh26it:
Thanks for the PR @5war00p!

Does this only check the entire content of the request to assume that it contains a URL?

We also have another plugin - https://github.com/Portkey-AI/gateway/blob/main/plugins/default/validUrls.ts

Is that conflicting?

5war00p:
This just does almost same, but instead of GET request I'm using HEAD request also we as we need to validate endpoint is reaching or not just header is fine no need validate dns separately.

narengogi (commented):
Looks good, just that this assumes that the response only contains a URL. in comparison,
https://github.com/Portkey-AI/gateway/blob/main/plugins/default/validUrls.ts feels more complete, I think it's a good idea to make a HEAD request to check the validity of a URL, you can perhaps add it as a fallback in the code of the existing plugin. That would be a 2 line change and more robust

5war00p:
To your point, `validUrl` ideally should be a regexp check thats what name speaks about. Where as `endpointReachability` make sure it reaches. If we really don't need 2 separate then we can merge, but validUrl can be used to validate a url input but if we make sure its a reachable one then merging both will be good idea.

plugins/default/endpointReachability.ts
```diff
@@ -0,0 +1,34 @@
...
+  _parameters,
+  eventType,
+  _options
+) => {
+  let error = null;
+  let verdict = false;
+  let data = null;
+
+  try {
+    let text = getText(context, eventType);
```
b4s36t4, line 27: My suggestion here is use regex to get all the URLs available from text and validate them using `URL` or some other way. 

With this we can have config weather to run plugin for all the endpoints or first endpoint found within the text.

What do you think? @5war00p

b4s36t4 (changes requested):

plugins/default/endpointReachability.ts
5war00p: Makes sense if text might contain >1 url, however this issue is to build a plugin for endpoint reachability right? and we should use this util func in another which extracts urls and check each url using this plugin. Correct me if I went off.

b4s36t4: Your thought makes sense, but this PR is a plugin PR which I think doesn't make sense to use it as utility function. I think we should focus on keep this as individual plugin. keeping that I prefer we should go for regex match. 

As an enhancement I think we shouldn't just blindly hit the URL, we should ask user which type of method they're thinking of ex: `GET` and `POST` and let plugin do `OPTIONS` request and make sure the endpoint(s) support the user requested methods.

let me know if this makes sense to you.

5war00p: I think if thats the case we should make it as a new separate plugin.

narengogi:
> To your point, `validUrl` ideally should be a regexp check thats what name speaks about. Where as `endpointReachability` make sure it reaches. If we really don't need 2 separate then we can merge, but validUrl can be used to validate a url input but if we make sure its a reachable one then merging both will be good idea.

sorry for the late reply @5war00p , as you might have noticed during your testing, the plugins act on the request and response body, so your plugin assumes it recieves only the URL in the body. Hence the plugin is not of much use.

But it's a cool implementation like I said, but for now this cannot be merged

5war00p:
@narengogi  In that case we can do as @b4s36t4 suggested, extract links and verify endpoint reachability, but question is do we need to do that for all links in found in entire text and how we do we determine that?

narengogi:
@5war00p you can use regex to extract the URLs

VisargD:
Hey folks! Are we moving forward with this implementation?

VisargD:
Hey! Closing this PR because there has been no activity since November end. Please reopen it if you are still working on it.

PR #732, opened 2024-11-05
Adding model and provider reference routes

roh26it:
Added 2 new routes - `/v1/reference/models` and `/v1/reference/providers` that provide a reference to all supported providers and models on the AI gateway.

### DIFF
--- a/src/data/providers.json
+++ b/src/data/providers.json
@@ -1,307 +1,307 @@
 {
-    "object": "list",
-    "version": "1.0.0",
-    "data": [
-        {
-            "id": "ai21",
-            "name": "AI21",
-            "object": "provider",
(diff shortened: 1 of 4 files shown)


src/data/providers.json
```diff
@@ -0,0 +1,307 @@
+{
+  "object": "list",
+  "version": "1.0.0",
+  "data": [
+    {
+      "id": "ai21",
```
narengogi, line 6: would recommend using the enums defined in `/src/globals` for all the providerIds in this file and in `models.json`

roh26it: Fair, but since this is loaded from a json file we cannot use enums in it.

narengogi (approved):

Issue #680, opened 2024-10-15
[SDK]: Vertex Controlled Output with Pydantic


b4s36t4:
### What Happened?

Vertex controlled output generation with Pydantic models when using with SDK

PR #733, opened 2024-11-05
fix: zod validation for vertex

b4s36t4:
**Title:** 
- Fix zod validation for vertex.

**Motivation:** (optional)

- Part of #680


narengogi (approved):
would recommend adding a little more context and description to the PR or in a GitHub issue

PR #743, opened 2024-11-12
Support shorthand format for guardrails in config

roh26it:


src/handlers/handlerUtils.ts
```diff
@@ -784,6 +810,16 @@ export async function tryTargetsRecursively(
     currentTarget.requestTimeout = inheritedConfig.requestTimeout;
   }
 
+  if (currentTarget.inputGuardrails) {
```
VisargD, line 525: Should we also handle scenario where user might have sent both: before_request_hooks and input_guardrails?

.gitignore
```diff
@@ -141,3 +141,4 @@ build
 .idea
 plugins/**/.creds.json
 plugins/**/creds.json
+src/handlers/test.ts
```
narengogi: is this intentional?

roh26it:
We should also support ids as part of this. Will need to add that to the Portkey middleware

VisargD (dismissed):

.gitignore
VisargD: Reverted this change

Issue #664, opened 2024-10-07
[Feature] Support OpenAI realtime API


narengogi:
### What Would You Like to See with the Gateway?

https://platform.openai.com/docs/guides/realtime?text-generation-quickstart-example=text

Ran-Mewo:
Could support for `gpt-4o-audio-preview` be also added?

PR #758, opened 2024-11-21
Feat: openai and azure-openai realtime api support

VisargD:
Title:

Support for OpenAI realtime API
Related Issues: (optional)

https://github.com/Portkey-AI/gateway/issues/664


src/handlers/realtimeHandler.ts
```diff
@@ -0,0 +1,75 @@
+import { Context } from 'hono';
```
narengogi, line 1: we should add exception handling here as well

src/services/realtimeLLMEventParser.ts
```diff
@@ -0,0 +1,149 @@
+import { Context } from 'hono';
+
+export class RealTimeLLMEventParser {
```
narengogi: typo in classname

src/handlers/websocketUtils.ts
```diff
@@ -0,0 +1,77 @@
+import { Context } from 'hono';
```
narengogi, line 1: will delete the unused functions

narengogi (approved):

src/services/realtimeLLMEventParser.ts
VisargD: Done

Issue #770, opened 2024-11-27
Google transformer does not return error codes


narengogi:
### What Happened?

In the error transformer for google, error status is being mapped to code and not the code.
```
code: response.error.status ?? null
```

PR #771, opened 2024-11-27
fix error message transformation for google gemini

narengogi:
- #770


VisargD:
@narengogi - Is this also required for vertex?

VisargD (approved):

PR #789, opened 2024-12-03
add optional model name header for sagemaker

narengogi:
Add optional model name header for sagemaker


sk-portkey (dismissed):

sk-portkey:
@narengogi dont we need the same change in src/providers/sagemaker - SagemakerAPIConfig> headers?

sk-portkey (approved):

Issue #793, opened 2024-12-05
[Feature] Pangea Cloud Plugin


pangea-andrest:
### What Would You Like to See with the Gateway?

As part of Pangea Cloud we'd like to implement a plugin to call our service AI Guard (https://pangea.cloud/services/ai-guard/).
With AI Guard service users could eliminate PII, sensitive data, and malicious content from ingestion pipelines, LLM prompts and responses.
- Identify and redact various forms of PII or sensitive information in AI prompts, responses, and context RAG data.
- Uncover and block malware submitted by users or ingested through agents or RAG file ingestion.
- Remove links to malicious IP addresses, domains, and URLs embedded in prompts, responses, or data vectors.

PR #794, opened 2024-12-05
[Feature] Pangea Cloud Plugin

pangea-andrest:
**Title: Pangea Cloud Plugin** 
- Add a new plugin to call Pangea Cloud AI Guard service

**Description:** (optional)
- This plugin include a `before` and `after` request hooks to call AI Guard service to process LLM inputs and outputs to analyze and redact text to avoid manipulation of the model, addition of malicious content, and other undesirable data transfers.

**Motivation:** (optional)
With AI Guard service users could eliminate PII, sensitive data, and malicious content from ingestion pipelines, LLM prompts and responses.

Identify and redact various forms of PII or sensitive information in AI prompts, responses, and context RAG data.
Uncover and block malware submitted by users or ingested through agents or RAG file ingestion.
Remove links to malicious IP addresses, domains, and URLs embedded in prompts, responses, or data vectors.

- https://github.com/Portkey-AI/gateway/issues/793


VisargD:
Hey @pangea-andrest , thanks for the PR! We will review this soon.

plugins/pangea/manifest.json
```diff
@@ -0,0 +1,75 @@
...
+  "description": "Pangea Text Guard for scanning LLM inputs and outputs",
+  "credentials": {
+    "type": "object",
+    "properties": {
+      "token": {
+        "type": "string",
+        "label": "Pangea token",
+        "description": "AI Guard token. Get your token configured on Pangea User Console (https://pangea.cloud/docs/getting-started/configure-services/#configure-a-pangea-service).",
+        "encrypted": true
+      },
```
b4s36t4: Can we rename it to `apiKey` please, since that's the naming we're using for other plugins which will make it easier for others to easily switch to new plugin?

plugins/pangea/textGuard.ts
```diff
@@ -0,0 +1,56 @@
...
+  context: PluginContext,
+  parameters: PluginParameters,
+  eventType: HookEventType,
+  options: {}
+) => {
+  let error = null;
+  let verdict = false;
+  let data = null;
+  try {
+    const text = getText(context, eventType);
```
b4s36t4: This only works with `completions` request, which in this case could fail the request and can be confusing to the user. I believe pangea does support `chatCompletions` guard i.e `prompt-guard` as well. Maybe can you integrate that as well or return early if text is falsy.

plugins/pangea/textGuard.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    // TODO: Update to v1 once released
+    const url = `https://ai-guard.${parameters.credentials?.domain}/v1beta/text/guard`;
```
b4s36t4: Can we have a early check for the required fields please? i.e `recipe` and `domain` etc with meaningful errors please?

b4s36t4 (changes requested):
Hi, @pangea-andrest. Thanks for raising this PR. Things are almost good, can you please resolve the comments I have raised. 

Also you should be adding the plugin handler inside `plugins/index.ts`. 

```js
pangea: {
   textGuard: handler
}
```
Other than that it's looking good, tested and verified working fine!
cc: @VisargD

plugins/pangea/textGuard.ts
pangea-andrest: @b4s36t4 thanks a lot for your review. I've addressed your other comments but I'm missing this one. I'm not sure if I follow you, could you explain it a little bit more what would you like I implement and why it wouldn't work? You mean not all request has the field `text` set? What other fields should I check? 
I've just added the empty text error.

plugins/pangea/textGuard.ts
pangea-andrest: Added check to `domain` and `apiToken`, the other params are optionals.

plugins/pangea/manifest.json
pangea-andrest: Renamed!

plugins/pangea/textGuard.ts
b4s36t4: Thanks for the update, I think this should work. No updates needed.

plugins/pangea/textGuard.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (!parameters.credentials?.apiKey) {
+      throw Error(`'parameters.credentials.apiKey' must be set`);
+    }
+
+    // TODO: Update to v1 once released
+    const url = `https://ai-guard.${parameters.credentials.domain}/v1beta/text/guard`;
+
+    const text = getText(context, eventType);
+    if (!text) {
+      throw Error(`request or response text is empty`);
```
b4s36t4, line 33: this shouldn't block the whole request (i.e throwing error may close the request), instead you just return the plugin result with error field.

```
return {error: "request or response text is empty", verdict: true, data}
```

plugins/pangea/textGuard.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
```
b4s36t4: this shouldn't block the whole request (i.e throwing error may close the request), instead you just return the plugin result with error field.

return {error: "request or response text is empty", verdict: true, data}

b4s36t4 (changes requested):
Can you please fix this one, rest looks good to me. Will approve after this.

plugins/pangea/textGuard.ts
pangea-andrest: Updated. So we just want to return an error here but we do want to throw an error if there is missing required parameters as apiKey or domain, is that right?

b4s36t4: I don't suggest adding `Error` class here, you can simply put string for the `error` key itself. 

```javascript
return {error: "request or response text is empty", verdict: true, data}
```

pangea-andrest: Okey. Fixed now.

pangea-andrest:
@b4s36t4 @VisargD Please, let me know if we should do something else to get it ready. Thanks!

b4s36t4:
Hey, @pangea-andrest. Sorry missed it. Just a small change please, I think you forget to remove the `Error` class at two places. Can you verify it again pls?

plugins/pangea/textGuard.ts
```diff
@@ -0,0 +1,71 @@
...
+  let verdict = false;
+  let data = null;
+  try {
+    if (!parameters.credentials?.domain) {
+      throw Error(`'parameters.credentials.domain' must be set`);
+    }
+
+    if (!parameters.credentials?.apiKey) {
+      throw Error(`'parameters.credentials.apiKey' must be set`);
+    }
```
b4s36t4: Here, can remove it pls?

b4s36t4 (changes requested):

plugins/pangea/textGuard.ts
pangea-andrest: Sure, I'll do it. Just to confirm, I guess in this case we want to return `verdict: false`, right? I mean, we do not want that an error in credentials allow malicious content to pass it through.

b4s36t4: ~I don't think so, cause there's no guardrail action happening. It's a pure usage issue, we could simply retrun `verdict` as `true`.~

Yea, should be good to return `true`.

pangea-andrest: @b4s36t4 So, is it confirmed as `true`? It seems that after the edit, you might want to say `false`.

b4s36t4 (approved):
LGTM!

plugins/pangea/textGuard.ts
b4s36t4: It's ok, will be good for now.

Issue #738, opened 2024-11-08
[Feature] Enable importing this package's contents


noy4:
### What Would You Like to See with the Gateway?

I'd like to import contents like `app`, `HooksManager` from this package (currently not possible), and integrate it with my sveltekit app backend, instead of hosting this gateway as another backend server. Example:

```ts
// hooks.server.ts
import { app, HooksManager } from '@portkey-ai/gateway'
import { myPlugins } from './myPlugins'

const hooks = (c: Context, next: any) => {
  const hooksManager = new HooksManager(myPlugins);
  c.set('hooksManager', hooksManager);
  c.set('executeHooks', hooksManager.executeHooks.bind(hooksManager));
  return next();
};

app.use('*', hooks); // this middleware comes last... I don't think it's gonna work

export async function handle(input) {
  // ...

  if (url.startsWith('/v1'))
    return app.fetch(event.request, undefined, platform?.context)

  // ...
}
```

Currently, I need to put this package into my project repo as git submodule to import `app`, .etc. I'd be happy if there's no need to use git submodule.

### Hint
1. need to tweak `rollup.config.js`?
2. package.json will need this kind of stuff?
```json
{
  "main": "build/index.js",
  "module": "build/index.js",
  "types": "build/index.d.ts",
  "exports": {
    ".": {
      "types": "./build/index.d.ts",
      "import": "./build/index.js",
      "require": "./build/index.js"
    }
  }
}
```

narengogi:
@noy4 We are not so sure that we want to publish an officially maintained package though, because it would mean maintenance with regards to compatibility in different environments, we've found it best to use the gateway as a standalone application

narengogi:
@roh26it has test changes for this, we're not sure this is the approach we want to be taking
even this POC PR from @sandys deals with the same theme https://github.com/Portkey-AI/gateway/pull/1455

sandys:
@narengogi agreed on this. also Portkey is far more production tested - so it should be a server. However, my attempt is to load packages inside portkey so @noy4 usecase can be addressed

Issue #601, opened 2024-09-11
Using 'dns' package in 'validUrls' plugin breaks Edge Deployments


jpulec:
## Description
In edge environments, like on Vercel's edge, many node APIs are not supported and cause deployments to fail. Including `dns` in the `validUrls` plugin causes this to happen.

### What Should Have Happened?

Consider a different approach for DNS based url validation, or maybe find a way to better allow the code to be tree shaken in edge environments.

VisargD:
Hey @jpulec @noy4 - Sorry for the delay on this. We were trying to figure out a better way for cross-compatibility in general and lost track of this thread. We have removed the dns module dependency and replaced it with dns-over-http approach which should work fine in edge runtimes as well. Please let me know if you face issues with any other dependencies.

PR #795, opened 2024-12-08
improvement: inject plugins through HooksManager constructor

noy4:
Inject plugins through HooksManager constructor instead of directly define in it.

**Motivation:**
In our sveltekit project, we're using this repo as a submodule and import some part to create Hono instance and attach it to our backend. When we deploy it to Cloudflare Pages, it failed because the plugins in HooksManager depend on Node API (#601). So we copied `middlewares/hooks/index.ts` and commented out the plugins part, and use the hooks middleware in our Hono instance.

```ts
export class HookSpan { ... }

export class HooksManager {
  private spans: Record<string, HookSpan> = {}
  private plugins: any

  constructor() {
    // this.plugins = plugins // ←
  }
}

export const hooks = (c: Context, next: any) => {
  const hooksManager = new HooksManager();
  c.set('hooksManager', hooksManager);
  c.set('executeHooks', hooksManager.executeHooks.bind(hooksManager));
  return next();
};
```

if plugins are injectable through HooksManager constructor, we only need to copy the hooks middleware part.
```ts
export const hooks = (c: Context, next: any) => {
  const hooksManager = new HooksManager(plugins);
  // const hooksManager = new HooksManager(); // we will change here, pass nothing, or custom something
  c.set('hooksManager', hooksManager);
  c.set('executeHooks', hooksManager.executeHooks.bind(hooksManager));
  return next();
};
```

**Related Issues:**
- #738

### DIFF
--- a/src/middlewares/hooks/index.ts
+++ b/src/middlewares/hooks/index.ts
@@ -174,7 +174,7 @@ export class HooksManager {
   private spans: Record<string, HookSpan> = {};
   private plugins: any;
 
-  constructor() {
+  constructor(plugins?: any) {
     this.plugins = plugins;
   }
 
@@ -442,7 +442,7 @@ export class HooksManager {
 }
 
 export const hooks = (c: Context, next: any) => {
-  const hooksManager = new HooksManager();
+  const hooksManager = new HooksManager(plugins);
   c.set('hooksManager', hooksManager);
   c.set('executeHooks', hooksManager.executeHooks.bind(hooksManager));
   return next();


VisargD:
Hey, sorry for the delay. We are reviewing this.

VisargD:
Hey @noy4 , we have removed the node dependency from the plugin handler as mentioned in the original issue https://github.com/Portkey-AI/gateway/issues/601. Are you still facing any issues?

noy4:
@VisargD 
Not anymore. 

And we quit the described approach:
> using this repo as a submodule and import some part to create Hono instance and attach it to our backend.

We are still using this repo as submodule to setup auto deploy, but we decided to deploy gateway as it is, and added some features in our project side.
- hook the request and set provider info by model name
- hook the response ReadableStream and calculate the usage cost (with cost list we prepared)

So you can close this PR. Thank you for your help.

VisargD:
Got it, we are also adding cost and usage details this repo soon. Maybe that will eliminate all the custom handling that you are using. I will send more details regarding that once its available. Thanks!

Issue #823, opened 2024-12-18
[Feature] Support new openai developer messages role


narengogi:
### What Would You Like to See with the Gateway?

its the same as system role
https://platform.openai.com/docs/guides/text-generation#developer-messages
https://x.com/simonw/status/1869101725266932158

PR #824, opened 2024-12-18
support new openai developer role 

narengogi:
**Testing done:**
- Tested all updated providers

**Related Issues:**
- #823


VisargD:
Hey @narengogi - We also have a openai-base provider which is reused for all OpenAI compatible providers. We might also have to handle the new role because most of the providers have not yet announced the support for the new developer role.

Maybe we can add a setting called, isDeveloperRoleSupported in openai-base provider function and map it to system if its not true. Any thoughts?

VisargD (approved):

Issue #752, opened 2024-11-18
[Provider] Recraft AI


vrushankportkey:
https://www.recraft.ai/docs#generate-image

VisargD:
PR is merged!

PR #835, opened 2024-12-19
[Provider] Recraft AI

Ajay-Satish-01:
**Title:** 
- Add Recraft AI to Provider

**Description:** (optional)
- Add with image generation

- #752

### DIFF
--- a/src/providers/recraft-ai/imageGenerate.ts
+++ b/src/providers/recraft-ai/imageGenerate.ts
@@ -1,0 +1,84 @@
+import { RECRAFTAI } from '../../globals';
+import { ErrorResponse, ImageGenerateResponse, ProviderConfig } from '../types';
+
+export const RecraftAIImageGenerateConfig: ProviderConfig = {
+  prompt: {
+    param: 'prompt',
+    required: true,
+  },
+  style: {
+    param: 'style',
+    default: 'digital_illustration',
+  },
+  width: {
+    param: 'width',
+    default: 1024,
+    min: 256,
+    max: 2048,
+  },
+  height: {
+    param: 'height',
+    default: 1024,
+    min: 256,
+    max: 2048,
+  },
+  num_images: {
+    param: 'num_images',
+    default: 1,
+    min: 1,
+    max: 4,
+  },
+};
+
+interface RecraftApiResponse {
+  created: number;
+  data: { url: string }[];
+  error?: {
+    message: string;
+    type?: string;
+    param?: string;
+    code?: string;
+  };
+}
+
+export const RecraftAIImageGenerateResponseTransform = (
+  response: RecraftApiResponse,
+  responseStatus: number
+): ImageGenerateResponse | ErrorResponse => {
+  if (responseStatus !== 200 || response.error) {
+    const error = response.error || {
+      message: 'Unknown error occurred',
+      type: null,
+      param: null,
+      code: null,
+    };
+
+    return {
+      error: {
+        message: error.message,
+        type: error.type || null,
+        param: error.param || null,
+        code: error.code || null,
+      },
+      provider: RECRAFTAI,
+    };
+  }
+
+  if (!response.data || !Array.isArray(response.data)) {
+    return {
+      error: {
+        message: 'Invalid response format from Recraft API',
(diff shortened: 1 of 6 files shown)


src/providers/recraft-ai/imageGenerate.ts
```diff
@@ -0,0 +1,84 @@
...
+    min: 256,
+    max: 2048,
+  },
+  height: {
+    param: 'height',
+    default: 1024,
+    min: 256,
+    max: 2048,
+  },
+  num_images: {
```
VisargD: As per the Recraft docs, this parameter is called `n`.  Also, gateway provides a unified route for image generations where the parameters are accepted in OpenAI compatible format and then mapped to provider specific fields. 

So this mapping should have mapped `n`
```
n: {
param: 'n',
    default: 1,
    min: 1,
    max: 4,
  }
```

src/providers/recraft-ai/imageGenerate.ts
```diff
@@ -0,0 +1,84 @@
...
+export const RecraftAIImageGenerateConfig: ProviderConfig = {
+  prompt: {
+    param: 'prompt',
+    required: true,
+  },
+  style: {
+    param: 'style',
+    default: 'digital_illustration',
+  },
+  width: {
```
VisargD: width and height are not accepted by Recraft APIs. 
https://www.recraft.ai/docs#generate-image
Screenshot 2024-12-20 at 01 08 19

Ajay-Satish-01: Following other provider's format in code for image generation, should I keep it immutable for standardisation?

VisargD: Gateway follows OpenAI standard. So if any parameter can be mapped to OpenAI compatible parameter, it will accept it and then map it to provider specific parameter. You can check the existing. You can check other providers like segmind, lemonfox-ai, etc. which are present in the providers directory.

VisargD: Moreover, from Recraft docs, it looks like they are already OpenAI compatible. So you should be able to directly map all the fields

Ajay-Satish-01: Got it. Will makes the changes. Thanks

Ajay-Satish-01:
@VisargD according to the documentation:

`Note: style_id and style parameters are mutually exclusive. If neither of the two parameters is specified, the default style of realistic_image will be used`

I have included a commented block of code. Should I include that or will the included be fine?

VisargD:
> @VisargD according to the documentation:
> 
> `Note: style_id and style parameters are mutually exclusive. If neither of the two parameters is specified, the default style of realistic_image will be used`
> 
> I have included a commented block of code. Should I include that or will the included be fine?

We should let the provider itself take care of defaults and validation error. The commented code is not required. Gateway currently does not throw validation error for provider specific errors because maintaining parity would become a hassle.

Ajay-Satish-01:
> > @VisargD according to the documentation:
> > `Note: style_id and style parameters are mutually exclusive. If neither of the two parameters is specified, the default style of realistic_image will be used`
> > I have included a commented block of code. Should I include that or will the included be fine?
> 
> We should let the provider itself take care of defaults and validation error. The commented code is not required. Gateway currently does not throw validation error for provider specific errors because maintaining parity would become a hassle.

Removed commented code.

VisargD:
Please move the PR to ready state for review once its done.

VisargD:
Hey @Ajay-Satish-01 , following parameters are not added in the mapping:
- model - This is an OpenAI compatible param and can be mapped as it is.
- artistic_level and substyle - These are Recraft specific params which are not OpenAI compatible. But they can be added as it is in case user wants to use them. 

So we can add these 3 mapping in the `RecraftAIImageGenerateConfig`:
```
model: {
  param: 'model',
},
artistic_level: {
  param: 'artistic_level',
},
substyle: {
  param: 'substyle',
},
```

Everything else LGTM!

Ajay-Satish-01:
@VisargD I have added those parameters as well. Anything else that I should fix or add?

VisargD (approved):

PR #836, opened 2024-12-20
feature/reduce docker image size using build stage

unsync:
**Title:** 
By using a build stage and copying only the needed resources, we can reduce the image size from 1Gb to 137Mb (at least, on my local builds)

**Warning:** 
If the code owner can double check if any other local resources are needed in the docker image, that would be nice.


VisargD:
Hey @unsync - Is 1GB the compressed size that you are getting? The size of the latest image tag is 222 MB.

unsync:
> Hey @unsync - Is 1GB the compressed size that you are getting? The size of the latest image tag is 222 MB.

@VisargD uncompressed, checking the sizes locally.

Here is the diff from a bare repo :
Screenshot 2024-12-20 at 08 07 20

But, even we we put aside the size, i think using the build step + copying only the resources you need is a good practice, that can avoid ton include sensitive/unwanted/unneeded files in the docker image, wdyt ?

unsync:
You did such a great job for the build that compresses all sources into a single file, i just felt it was a waste of resources not to go all the way and make the smallest docker image possible ;)

VisargD:
Agree, I will run some tests on this. Thanks!

VisargD (approved):

Issue #837, opened 2024-12-22
Issues using tool calls with Google Gemini 


au-re:
### What Happened?

Hi all! I noticed some inconsistent behavior when switching models to google gemini regarding tool calls:

The same call works with either Anthropic or OpenAI but fails with Google. 

If `additionalProperties` is provided to the tool parameter definition we receive the error response below, although this is a valid parameter in the JSON Schema definition and is accepted by OpenAI.
```
400 Invalid response received from google: [{"error":{"code":400,"message":"Invalid JSON payload received. Unknown name \"additionalProperties\" at 'tools[0].function_declarations[0].parameters'
```

If `strict: true` is specified in the tool definition we get the following error:
```
400 Invalid response received from google: [{"error":{"code":400,"message":"Invalid JSON payload received. Unknown name \"strict\" at 'tools[0].function_declarations[0]'
``` 

Lastly even after removing these parameters we still get an error and the gateway crashes:
```
TypeError: Cannot read properties of undefined (reading 'parts')
```

Additional information: running version 1.8.5 of the gateway locally using node.

### What Should Have Happened?

1. Switching providers / model id should have been enough for making this request.
2. It should be possible to make tool calls to google gemini models

### Relevant Code Snippet

```ts
import { codeBlock } from "common-tags";
import { createHeaders } from "portkey-ai";
import OpenAI from "openai";

const openai = new OpenAI({
  apiKey: "...",
  baseURL: "...",
  defaultHeaders: createHeaders({
    provider: "google",
    apiKey: "...",
  }),
});

async function createChatCompletion() {
  const chat_complete = await openai.chat.completions.create({
    model: "gemini-1.5-flash",
    messages: [
      { role: "system", content: "You are a helpful assistant." },
      { role: "user", content: "..." },
    ],
    tool_choice: "auto",
    parallel_tool_calls: true,
    temperature: 0,
    stream: true,
    tools: [
      {
        type: "function",
        function: {
          name: "run_code",
          description: codeBlock`...`,
          parameters: {
            type: "object",
            properties: {
              foo: {
                type: "string",
                description: codeBlock`...`,
              },
              bar: {
                type: "string",
                description: "...",
              },
            },
            required: ["foo", "bar"],
          },
          // strict: true,
        },
      },
    ],
  });

  for await (const message of chat_complete) {
    console.log(JSON.stringify(message, null, 2));
  }
}

createChatCompletion();
```

narengogi:
[...]

PR #840, opened 2024-12-24
#837 delete unsupported fields in gemini models tool definitions

narengogi:
Tested by sending 
- `strict` inside tools[0].function
- `additional_properties` inside tools[0].function.parameters.properties

gemini schema: https://cloud.google.com/vertex-ai/docs/reference/rest/v1/Tool
openai schema: https://platform.openai.com/docs/api-reference/chat/create#chat-create-tools

- #837


csgulati09 (dismissed):

src/providers/google-vertex-ai/chatComplete.ts
```diff
@@ -269,6 +269,10 @@ export const VertexGoogleChatCompleteConfig: ProviderConfig = {
       const tools: any = [];
       params.tools?.forEach((tool) => {
         if (tool.type === 'function') {
+          // these are not supported by google
+          delete tool.function?.parameters?.properties?.additional_properties;
```
VisargD: I think its also possible to have additional_properties in nested object. So we would have to handle that as well. Can you please verify once?

narengogi: done!

VisargD (approved):

Issue #819, opened 2024-12-17
[Feature] Support batches and files as unified API routes


narengogi:
### What Would You Like to See with the Gateway?

Requirements:
- [ ] Support /v1/files routes as a unified API
- [ ] Support /v1/batches routes as a unified API
- [ ] Additionally support a route to get batches output. something like /v1/batches/:batchId/output

This would require some considerable amount of changes in the gateway, so for v1
1. support the following providers: openai, bedrock, cohere
2. bedrock would require uploading to s3 as chunks as streaming a file to s3 without knowing the content-length is not allowed, please make provisions for that
3. File uploads should not increase memory consumption in worker environments. even when transforming the files, only 100kb chunks should be read

anterart:
Hi @narengogi, when do you think this feature will be available?
I'm really looking up to it :)
Also is Google's Vertex AI & Azure OpenAI's batch API will be supported?

VisargD:
Hey @anterart! We have merged the PR. It will officially be a part of release happening today

narengogi:
batches as a unified route is available, docs at https://portkey.ai/docs/product/ai-gateway/batches#batches

@anterart I have to add support for vertex and azure, transforming files on the fly without loading into memory is a bit of tricky work, took me some time to implement for bedrock

PR #862, opened 2025-01-13
Feature: Files and Batches as a unified route

narengogi:
- #819 

Feature to add batches and files as a unified route
This add the following routes:

POST `/v1/files`
GET `/v1/files`
GET `/v1/files/:fileId`
DELETE `/v1/files/:fileId`

POST `/v1/batches`
GET `/v1/batches`
GET `/v1/batches/:batchId`
POST `/v1/batches/:batchId/cancel`

✨ Gateway specific additional route for convenience (Not in OpenI spec) ✨
GET `/v1/batches/:batchId/output`

**Description**
- For file uploads, the gateway needs to transform files to the provider format (like bedrock) on the fly, the files can be as large as 20GB, this implementation does not load the files into memory, rather it transforms the data streams on the fly with nodeJS streams and pushes to the other endpoint with a half duplex connection. 

- Similar streaming transformers for getting batch output and getting files is implemented, where the file stream is transformed on the fly

- I've had to upgrade wrangler version for nodeJS compatibility to add Polyfill support for nodeJS crypto module which is used for signing the AWS requests, although the gateway has `@smithy/signature-v4` package installed, I wasn't able to get it to work for the s3 file upload routes, I'll have to revisit this

**Testing done**
- To make sure that we're not adding any additional memory overhead, and that there are no memory leaks, I've monitored the memory consumption with the `top` command on my machine
image
during testing a 1gb file upload, memory went up by about 40MiB, but returned to normal shortly after

**TO-DO for later**
- Revisit and simplify the AWS signer code, we should be able to use the `@smithy/signature-v4` package without having to write all the signing code ourselves, or alternatively we can remove the smithy signing package and implement the functions ourselves


src/providers/cohere/getBatchOutput.ts
VisargD, line 1: Can we remove this file? Everything is commented.

src/providers/ai21/api.ts
```diff
@@ -8,7 +8,7 @@ const AI21APIConfig: ProviderAPIConfig = {
     };
     return headers;
   },
-  getEndpoint: ({ fn, gatewayRequestBody }) => {
+  getEndpoint: ({ fn, gatewayRequestBodyJSON: gatewayRequestBody }) => {
```
VisargD: Nitpick: This might become a little bit confusing. Both gatewayRequestBody and gatewayRequestBodyJSON are passed to the getEndpoint method. But then we are again renaming it during destructuring.

narengogi: renamed

src/providers/bedrock/index.ts
```diff
@@ -36,142 +36,186 @@ import {
...
 const BedrockConfig: ProviderConfigs = {
   api: BedrockAPIConfig,
+  requestHandlers: {
+    uploadFile: BedrockUploadFileRequestHandler,
+    retrieveFile: BedrockRetrieveFileRequestHandler,
+    getBatchOutput: BedrockGetBatchOutputRequestHandler,
+    retrieveFileContent: BedrockRetrieveFileContentRequestHandler,
+  },
   getConfig: (params: Params) => {
-    if (!params.model) {
```
VisargD, line 55: We have removed this check completely in this update. Please add it back. We can keep a check for chatComplete and complete request types

src/providers/bedrock/api.ts
```diff
@@ -56,25 +110,34 @@ const BedrockAPIConfig: ProviderAPIConfig = {
...
+      service,
       providerOptions.awsRegion || '',
       providerOptions.awsAccessKeyId || '',
       providerOptions.awsSecretAccessKey || '',
       providerOptions.awsSessionToken || ''
     );
   },
-  getEndpoint: ({ fn, gatewayRequestBody }) => {
-    const { model, stream } = gatewayRequestBody;
-    if (!model) throw new GatewayError('Model is required');
```
VisargD, line 69: This check has been removed in this update. We should keep it for chatComplete and complete request types atleast

VisargD (dismissed):

src/providers/bedrock/index.ts
narengogi: I've added the check `if (params.model) {` on line 84, this has the same effect

src/providers/bedrock/api.ts
narengogi: done!

Issue #866, opened 2025-01-14
[Feature] Support milvus db


narengogi:
### What Would You Like to See with the Gateway?

this can be a simple proxy route support which allows headers and request modification

PR #867, opened 2025-01-14
support milvus db

narengogi:
- #866

### DIFF
--- a/src/handlers/handlerUtils.ts
+++ b/src/handlers/handlerUtils.ts
@@ -93,9 +93,9 @@ export function constructRequest(
   // Add any headers that the model might need
   headers = {
     ...baseHeaders,
+    ...(fn === 'proxy' && proxyHeaders),
     ...headers,
     ...forwardHeadersMap,
-    ...(fn === 'proxy' && proxyHeaders),
   };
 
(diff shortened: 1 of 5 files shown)


src/handlers/handlerUtils.ts
```diff
@@ -93,9 +93,9 @@ export function constructRequest(
   // Add any headers that the model might need
   headers = {
     ...baseHeaders,
+    ...(fn === 'proxy' && proxyHeaders),
```
VisargD: Hey, is there any reason to change this order in the PR? Not sure if this is in the scope of this integration.

narengogi: This integration is using the proxy route while using `milvus/api.ts` for constructing the headers. This is causing the Authorization header that is set in the `getHeaders()` method to get overriden with the one passed during the API call.

If we want the user to explicitly pass `authorization` header this way `Authorization: Bearer abcasd` then I can remove this change, but if we want to support `Authorization: abcasd` and also `Authorization: Bearer abcasd`, then this change is required

narengogi: reverting this change for now

VisargD (approved):

PR #870, opened 2025-01-17
Feat: Mutator Hooks

### LINKED ISSUE #872: [Feature] Mutator Hooks
Gateway currently allow hooks of type `guardrail`. These hooks perform various checks and validations and then return a boolean verdict. But there are various use cases where the context of request needs to be transformed based on hook results. In order to achieve this, we will be introducing a new type of hook called `mutator`.

Mutator hooks will have the capability of transforming the request and response instead of just passing a verdict. These hooks can then be used for various use cases like PII redaction, PHI redaction, Regex based redaction, etc. And these hooks can also be further extended to implement custom logic like abstracting out the configs and provider configs. 

For the first release, we will be implementing it for the redaction use case for existing plugins like portkey.pii, pangea.pii, etc. But the hooks middleware itself will be capable of mutating any other parts of the body and headers.

VisargD:
**Title:** 
- Mutator Hooks

- Closes #872

### DIFF
--- a/plugins/index.ts
+++ b/plugins/index.ts
@@ -34,6 +34,13 @@ import { handler as patronustoxicity } from './patronus/toxicity';
 import { handler as patronuscustom } from './patronus/custom';
 import { mistralGuardrailHandler } from './mistral';
 import { handler as pangeatextGuard } from './pangea/textGuard';
+import { handler as portkeyredactPii } from './portkey/redactPii';
+import { handler as promptfooRedactPii } from './promptfoo/redactPii';
+import { handler as promptfooHarm } from './promptfoo/harm';
+import { handler as promptfooGuard } from './promptfoo/guard';
+import { handler as pangearedactPii } from './pangea/redactPii';
+import { handler as patronusredactPii } from './patronus/redactPii';
+import { handler as patronusredactPhi } from './patronus/redactPhi';
 
 export const plugins = {
   default: {
@@ -58,6 +65,7 @@ export const plugins = {
     language: portkeylanguage,
     pii: portkeypii,
     gibberish: portkeygibberish,
+    redactPii: portkeyredactPii,
   },
   aporia: {
     validateProject: aporiavalidateProject,
@@ -81,11 +89,19 @@ export const plugins = {
     retrievalAnswerRelevance: patronusretrievalAnswerRelevance,
     toxicity: patronustoxicity,
     custom: patronuscustom,
+    redactPii: patronusredactPii,
+    redactPhi: patronusredactPhi,
   },
   mistral: {
     moderateContent: mistralGuardrailHandler,
   },
   pangea: {
     textGuard: pangeatextGuard,
+    redactPii: pangearedactPii,
+  },
+  promptfoo: {
+    redactPii: promptfooRedactPii,
+    harm: promptfooHarm,
+    guard: promptfooGuard,
   },
 };
--- a/plugins/patronus/redactPhi.ts
+++ b/plugins/patronus/redactPhi.ts
@@ -24,8 +24,7 @@ const redactPhi = async (text: string, credentials: any) => {
 
   const positionsData = evalResult.evaluation_result.additional_info;
   if (
-    positionsData &&
-    positionsData.positions &&
+    Array.isArray(positionsData?.positions) &&
     positionsData.positions.length > 0
   ) {
     const longestPosition = findAllLongestPositions(positionsData);
--- a/plugins/patronus/redactPhi.ts
+++ b/plugins/patronus/redactPhi.ts
@@ -1,0 +1,95 @@
+import {
+  HookEventType,
+  PluginContext,
+  PluginHandler,
+  PluginParameters,
+} from '../types';
+import { getCurrentContentPart, setCurrentContentPart } from '../utils';
+import { findAllLongestPositions, postPatronus } from './globals';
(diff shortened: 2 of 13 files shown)


narengogi (dismissed):

plugins/index.ts
```diff
@@ -81,11 +89,19 @@ export const plugins = {
     retrievalAnswerRelevance: patronusretrievalAnswerRelevance,
     toxicity: patronustoxicity,
     custom: patronuscustom,
+    redactPii: patronusredactPii,
+    redactPhi: patronusredactPhi,
```
b4s36t4, line 93: Now that `mutators` are becoming standard we can make these keys as const/enum for better reading and maintainability.

plugins/patronus/redactPhi.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  const evaluationBody: any = {
+    output: text,
+  };
+
+  const result: any = await postPatronus(
+    evaluator,
+    credentials,
+    evaluationBody
+  );
```
b4s36t4, line 22: We're not validating the `credentials`.

b4s36t4: `try/catch` is needed here, post call throws error.

plugins/patronus/redactPhi.ts
```diff
@@ -0,0 +1,95 @@
...
+  const evalResult = result.results[0];
+
+  const positionsData = evalResult.evaluation_result.additional_info;
+  if (
+    positionsData &&
+    positionsData.positions &&
+    positionsData.positions.length > 0
+  ) {
```
b4s36t4: better to add `Array.isArray` check here just to be safe.

plugins/patronus/redactPii.ts
```diff
@@ -0,0 +1,134 @@
...
+      error: null,
+      verdict: true,
+      data:
+        transformedText.filter((text) => text !== null).length > 0
+          ? { flagged: true }
+          : null,
+      transformedData,
+    };
+  } catch (e: any) {
+    delete e.stack;
```
b4s36t4, line 125: any specific reason to delete the `stack` here?

src/handlers/handlerUtils.ts
```diff
@@ -192,10 +192,14 @@ export function selectProviderByWeight(providers: Options[]): Options {
   throw new Error('No provider selected, please check the weights');
 }
 
-export function convertGuardrailsShorthand(guardrailsArr: any, type: string) {
-  return guardrailsArr.map((guardrails: any) => {
+export function convertHooksShorthand(
+  hooksArr: any,
+  type: string,
+  hookType: string
```
b4s36t4: can be `HookType` right, instead of string.?

src/handlers/handlerUtils.ts
```diff
@@ -586,14 +596,40 @@ export async function tryTargetsRecursively(
...
+        'output',
+        'guardrail'
+      ),
+    ];
+  }
+
+  if (currentTarget.inputMutators) {
+    currentTarget.beforeRequestHooks = [
+      ...(currentTarget.beforeRequestHooks || []),
+      ...convertHooksShorthand(currentTarget.inputMutators, 'input', 'mutator'),
```
b4s36t4: Maybe we could use the enum here `HookType.MUTATOR`?

b4s36t4 (changes requested):

plugins/patronus/redactPhi.ts
VisargD: The handler itself catches the error that comes from the reused `post` function. And the message is then added as error property in handler response.

plugins/index.ts
VisargD: Can you please elaborate? Not able to understand the requirement. 

Ideally, we want `plugins/index.ts` as just a handler map. Users can remove/add their own handler mapping before deployment. We do not plan to add any other metadata in this map apart from the handler mapping.

plugins/patronus/redactPhi.ts
VisargD: Agree, have made the change

plugins/patronus/redactPii.ts
VisargD: For partner plugins, its usually the choice of the partner themselves to write the handler. In this case, patronus team wanted the error message to be without stack.

But I think its better to streamline this in future. Will raise to streamline all the plugin responses.

src/handlers/handlerUtils.ts
VisargD: It makes sense. I wanted to keep the middleware isolated in themselves. No imports should be made from a middleware in the main code. But we are not following it fully till now so I will make the change. 

Will plan out a better standard for imports from middlewares.

VisargD: Made the change

src/handlers/handlerUtils.ts
VisargD: Made the change. 

And I will address another issue in a separate PR (as mentioned in other comment):
 https://github.com/Portkey-AI/gateway/pull/870#discussion_r1922133195

plugins/patronus/redactPhi.ts
b4s36t4: oh, sorry. Thought it was in handler itself. Can be closed.

b4s36t4 (approved):
LGTM!

PR #879, opened 2025-01-20
Feat/bedrock guardrails

### LINKED ISSUE #859: [Plugin] AWS Bedrock Guardrails
Bedrock has sophisticated guardrails for various use cases and has a separate API that can be invoked to run the creaed Guardrails. Having an integration with Bedrock Guardrails would enable us to cater to users who have majority of their workloads on AWS and want to use Portkey along with it.

https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails-use-independent-api.html
https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails-how.html

b4s36t4:
**Title:** 
- Adds bedrock guardrail support

- /closes #859

### DIFF
--- a/plugins/bedrock/index.ts
+++ b/plugins/bedrock/index.ts
@@ -1,6 +1,6 @@
-import { generateAWSHeaders } from '../../src/providers/bedrock/utils';
 import { PluginHandler } from '../types';
 import { getText, HttpError, post } from '../utils';
+import { generateAWSHeaders } from './util';
 
 const REQUIRED_CREDENTIAL_KEYS = [
   'accessKeyId',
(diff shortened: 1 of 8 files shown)


b4s36t4:
Screenshot 2025-01-21 at 6 40 53 PM

plugins/bedrock/index.ts
```diff
@@ -0,0 +1,211 @@
+import { generateAWSHeaders } from '../../src/providers/bedrock/utils';
```
VisargD: Plugins are ideally supposed to be isolated in nature. So it is not feasible to import modules from the gateway code. They should not have any other dependencies. 

If we want the functions, its better to write it in plugin provider util itself. In short term, not importing module might produce redundant code. But in long term, this will help in better structuring in case we want to separate out plugins in a separate repo or projects.

b4s36t4: Done!

VisargD (approved):

PR #534, opened 2024-08-21
Feat/azure inference support

b4s36t4:
**Title:** 
- Added support to communicate with azure inference API

**Related Issues:** (optional)
- #492

PR #901, opened 2025-01-29
chore: allow azure-ai to accept the extra-params from the headers

b4s36t4:
ref https://github.com/Portkey-AI/gateway/pull/534/files/c5b5cd397957e65fa4de2327d49870dbc1cf0bba#r1933769819

doc: https://learn.microsoft.com/en-us/azure/ai-foundry/model-inference/reference/reference-model-inference-chat-completions#request-header

### DIFF
--- a/src/providers/azure-ai-inference/api.ts
+++ b/src/providers/azure-ai-inference/api.ts
@@ -20,11 +20,15 @@ const AzureAIInferenceAPI: ProviderAPIConfig = {
     return `https://${azureEndpointName}.${azureRegion}.inference.ml.azure.com/score`;
   },
   headers: ({ providerOptions }) => {
-    const { apiKey, azureDeploymentType, azureDeploymentName } =
-      providerOptions;
+    const {
+      apiKey,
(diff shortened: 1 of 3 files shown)


src/providers/azure-ai-inference/api.ts
```diff
@@ -20,11 +20,15 @@ const AzureAIInferenceAPI: ProviderAPIConfig = {
...
+    const {
+      apiKey,
+      azureDeploymentType,
+      azureDeploymentName,
+      azureExtraParams,
+    } = providerOptions;
     const headers: Record<string, string> = {
       Authorization: `Bearer ${apiKey}`,
-      'extra-parameters': 'ignore',
+      'extra-parameters': azureExtraParams || 'pass-through',
```
VisargD, line 31: We are changing the existing behaviour here.

b4s36t4: Yea, azure itself broke it first 😅. So it's not a problem I think.

PR #913, opened 2025-02-03
Acuvity-portkey integration.

abhijitherekar:
**Initial PR for acuvity portkey integration.:** 

**Description:**
- Add acuvity guardrails support into the portkey gateway.


package.json
```diff
@@ -63,6 +63,7 @@
     "@types/ws": "^8.5.12",
     "husky": "^9.1.4",
     "jest": "^29.7.0",
+    "jose": "^5.9.6",
```
narengogi: hey @abhijitherekar, I know this is still draft, but I would not recommend adding an external library for just jwt decoding, you can write a function and put it in utils

we try to keep the portkey build folder small, currently it's around ~500kb

plugins/acuvity/helper.ts
```diff
@@ -0,0 +1,223 @@
+import { Extraction, Textualdetection, TextualdetectionType } from './model.js';
```
VisargD: Can you please remove extension from the import?

package.json
abhijitherekar: done.

plugins/acuvity/helper.ts
abhijitherekar: done.

plugins/acuvity/helper.test.ts
```diff
@@ -0,0 +1,120 @@
...
+      // Arrange
+      const extraction: Extraction = {
+        exploits: {
+          PROMPT_INJECTION: 0.8,
+        },
+      };
+      const threshold = 0.7;
+
+      // Act
+      const result = responseHelper.evaluate(
```
VisargD, line 22: Hey, all the test cases are written for responseHelper class. Ideally, we would want to have the test cases for individual top level handler. Can you please also add tests for `handler` of scan.ts?

You can check other providers like portkey, bedrock, etc. for reference.

plugins/acuvity/scan.ts
```diff
@@ -0,0 +1,259 @@
...
+
+function evaluateAllParameters(
+  extraction: any,
+  parameters: PluginParameters,
+  responseHelper: ResponseHelper
+): Array<{ parameter: string; result: GuardResult }> {
+  const results: Array<{ parameter: string; result: GuardResult }> = [];
+
+  // Check prompt injection
+  if (parameters.prompt_injection.enabled) {
```
VisargD: Is it possible to make these as top level parameters like `parameters.prompt_injection: true/false` and `parameters.prompt_injection_threshold: 0.5`. We use a json driven UI builder to render these params in the product. And nested parameters are not currently supported by UI.

plugins/acuvity/helper.test.ts
abhijitherekar: done.

narengogi:
Damn, that's so fully featured 🤵🏽‍♂️, reviewing it now

VisargD (dismissed):

abhijitherekar:
@VisargD , I have updated the branch. Now its in sync with the main. 
Please, can you re-approve the PR.

plugins/acuvity/scan.ts
```diff
@@ -0,0 +1,285 @@
...
+    const extractionResult = result as { extractions: Array<{ data: string }> };
+    const respTextArray = extractionResult.extractions.map(
+      (extraction) => extraction.data
+    );
+    let guardResults = new Set();
+
+    // Loop through all extractions
+    for (const extraction of result.extractions) {
+      // Evaluate parameters for current extraction
+      guardResults = evaluateAllParameters(
```
VisargD: Hey, here we keep overwriting the guardResults instead of add new items to the set. Is that expected?

plugins/acuvity/scan.ts
```diff
@@ -0,0 +1,285 @@
...
+    // if yes then return the redacted data with verdict = true.
+    // else verdict = false, as we found other detections.
+    if (
+      guardResults.size == 1 &&
+      redactionList.length > 0 &&
+      guardResults.has(GuardName.PII_DETECTOR)
+    ) {
+      verdict = true;
+      setCurrentContentPart(context, eventType, transformedData, respTextArray);
+      transformed = true;
```
VisargD: setCurrentContentPart should be user regardless of whether PII was the only thing detected.

From what I understand, here we are only redacting the data if PII was the only thing that was present in the result. If anything else like harmfulContent is found, we will not perform redaction at all even if its possible.

abhijitherekar: fixed it in the latest code. thanks for catching this, I wasn't aware of the deny part is optional.

VisargD (approved):

abhijitherekar:
Hi @VisargD , please lets merge this if you are good. 

Thanks

PR #932, opened 2025-02-08
handle tool_calls mapping in gemini responses when there is one part tool call and one part text

narengogi:
#933

### DIFF
--- a/src/providers/google-vertex-ai/api.ts
+++ b/src/providers/google-vertex-ai/api.ts
@@ -2,13 +2,9 @@ import { Options } from '../../types/requestBody';
 import { endpointStrings, ProviderAPIConfig } from '../types';
 import { getModelAndProvider, getAccessToken } from './utils';
 
-const shouldUseBeta1Version = (provider: string, inputModel: string) => {
-  if (
-    provider === 'meta' ||
-    inputModel.includes('gemini-2.0-flash-thinking-exp')
-  )
-    return true;
-  return false;
+const getApiVersion = (provider: string, inputModel: string) => {
+  if (provider === 'meta') return 'v1beta1';
+  return 'v1';
 };
 
 const getProjectRoute = (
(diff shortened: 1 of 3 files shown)


narengogi:
I'm keeping this in draft because I'm not able to test gemini thinking models responses, the thinking models are not returning the CoT message when invoked directly using vertex api endpoints either

src/providers/google-vertex-ai/chatComplete.ts
```diff
@@ -663,44 +664,35 @@ export const GoogleChatCompleteResponseTransform: (
       provider: GOOGLE_VERTEX_AI,
       choices:
         response.candidates?.map((generation, index) => {
-          const containsChainOfThoughtMessage =
```
VisargD, line 666: Is this a merge mistake? Not sure why are we removing thinking model support

src/providers/google-vertex-ai/api.ts
```diff
@@ -26,10 +22,7 @@ const getProjectRoute = (
   }
 
   const { provider } = getModelAndProvider(inputModel as string);
-  let routeVersion = provider === 'meta' ? 'v1beta1' : 'v1';
-  if (shouldUseBeta1Version(provider, inputModel)) {
```
VisargD, line 30: Same problem here as well. This will remove the thinking model support

narengogi: thinking model is now supported in v1 endpoint

src/providers/google-vertex-ai/chatComplete.ts
narengogi: thinking model is supported, I've modified the code a bit

VisargD (dismissed):

VisargD:
@narengogi  TODO: Need to check for streaming as well. Not a blocker for this PR

src/providers/google-vertex-ai/chatComplete.ts
```diff
@@ -663,44 +664,35 @@ export const GoogleChatCompleteResponseTransform: (
...
-                content = generation.content.parts[1]?.text;
-              else
-                content =
-                  generation.content.parts[0]?.text +
-                  '\r\n\r\n' +
-                  generation.content.parts[1]?.text;
+          // transform tool calls and content by iterating over the content parts
+          let toolCalls: ToolCall[] = [];
+          let content: string | undefined;
+          for (const part of generation.content?.parts) {
```
VisargD: This will fail if parts is not present (which is expected because we are using optional chaining)

PR #945, opened 2025-02-17
improvement: handle azure workload identity authentication

abestel:
**Title:** 
- Handle azure workload identity authentication

**Description:** (optional)
So far the Azure OpenAI integration was handling authentication using Client ID / Client Secret and Managed identity using the IMDS endpoint which is deprecated in favor of Workload Identity (using the public OAuth2 endpoint of Entra ID).

This changeset aims at handling this new authentication type.

Note that this requires reading environment variables set by the Azure runtime onto the virtual machine / pod using a workload identity. It also needs to read a file on disk (containing an assertion to use to exchange against a JWT).

**Motivation:** (optional)
- managed identity authentication using IMDS is deprecated and should be replaced by Workload Identity authentication

/

### DIFF
--- a/src/middlewares/requestValidator/schema/config.ts
+++ b/src/middlewares/requestValidator/schema/config.ts
@@ -4,6 +4,7 @@ import {
   VALID_PROVIDERS,
   GOOGLE_VERTEX_AI,
   TRITON,
+  AZURE_OPEN_AI,
 } from '../../../globals';
 
 export const configSchema: any = z
@@ -107,6 +108,7 @@ export const configSchema: any = z
     openai_organization: z.string().optional(),
     // AzureOpenAI specific
     azure_model_name: z.string().optional(),
+    azure_auth_mode: z.string().optional(),
     strict_open_ai_compliance: z.boolean().optional(),
   })
   .refine(
@@ -123,6 +125,8 @@ export const configSchema: any = z
         (value.vertex_service_account_json || value.vertex_project_id);
       const hasAWSDetails =
         value.aws_access_key_id && value.aws_secret_access_key;
+      const hasAzureAuth =
+        value.provider == AZURE_OPEN_AI && value.azure_auth_mode;
 
       return (
         hasProviderApiKey ||
@@ -137,7 +141,8 @@ export const configSchema: any = z
         value.after_request_hooks ||
         value.before_request_hooks ||
         value.input_guardrails ||
-        value.output_guardrails
+        value.output_guardrails ||
+        hasAzureAuth
       );
     },
     {
--- a/src/providers/azure-openai/api.ts
+++ b/src/providers/azure-openai/api.ts
@@ -2,14 +2,17 @@ import { ProviderAPIConfig } from '../types';
 import {
   getAccessTokenFromEntraId,
   getAzureManagedIdentityToken,
+  getAzureWorkloadIdentityToken,
 } from './utils';
+import { env } from 'hono/adapter';
+import fs from 'fs';
 
 const AzureOpenAIAPIConfig: ProviderAPIConfig = {
   getBaseURL: ({ providerOptions }) => {
     const { resourceName } = providerOptions;
     return `https://${resourceName}.openai.azure.com/openai`;
   },
-  headers: async ({ providerOptions, fn }) => {
+  headers: async ({ c, providerOptions, fn }) => {
     const { apiKey, azureAuthMode } = providerOptions;
 
     if (azureAuthMode === 'entra') {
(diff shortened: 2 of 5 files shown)


src/middlewares/requestValidator/schema/config.ts
```diff
@@ -123,6 +125,8 @@ export const configSchema: any = z
         (value.vertex_service_account_json || value.vertex_project_id);
       const hasAWSDetails =
         value.aws_access_key_id && value.aws_secret_access_key;
+      const hasAzureAuth =
```
abestel, line 128: I don't want this PR to get too big, but we could be doing more checks here, based on the authentication mode, some parameters may be required, and we could fail early.

For example, if `authMode` is `entra`, the `tenantId`, `clientId` and `clientSecret` would be required as well.

For now, this allows to validate the configuration for Azure OpenAI when an API Key is not required because we use another form of authentication.

By the way, I think this was not working before, the config would be rejected.

src/providers/azure-openai/api.ts
```diff
@@ -2,14 +2,17 @@ import { ProviderAPIConfig } from '../types';
 import {
   getAccessTokenFromEntraId,
   getAzureManagedIdentityToken,
+  getAzureWorkloadIdentityToken,
 } from './utils';
+import { env } from 'hono/adapter';
+import fs from 'fs';
```
abestel: Note that I'm not an expert of Hono.
This is most likely very specific to the node runtime, not sure if we need to specify it in any way here.
Definitely open to feedback.

src/providers/azure-openai/api.ts
```diff
@@ -39,6 +42,32 @@ const AzureOpenAIAPIConfig: ProviderAPIConfig = {
         Authorization: `Bearer ${accessToken}`,
       };
     }
+    if (azureAuthMode === 'workload') {
+      const { azureWorkloadClientId } = providerOptions;
+
+      const authorityHost = env(c).AZURE_AUTHORITY_HOST;
+      const tenantId = env(c).AZURE_TENANT_ID;
+      const clientId = azureWorkloadClientId || env(c).AZURE_CLIENT_ID;
+      const federatedTokenFile = env(c).AZURE_FEDERATED_TOKEN_FILE;
```
abestel, line 50: All these are well known environment variables from azure workload identity.

narengogi:
hey @abestel Since you're using `fs`, this would work well in nodeJS environment, but not cloudflare workers, but that's fine, let me just test these changes in both the environments and go through the code 👨‍💻

abestel:
Hey @narengogi 
FYI, I rebased the branch with the latest version of main because some conflicts appeared. There was no change in the code itself.

narengogi:
@abestel We'll test this today

src/middlewares/requestValidator/schema/config.ts
narengogi: @abestel this key is redundant, in `azure-openai/api` you could just check if `azureWorkloadClientId` is present to use workload identity based authentication

please remove this, rest of the code looks good to go

narengogi (changes requested):
added minor comment, rest looks good

src/middlewares/requestValidator/schema/config.ts
abestel: I’m not sure I understand your point.

The key `azureWorkloadClientId` is optional, even when using the workload identity authentication mode. It is only required when a resource is associated to multiple service principals, but by default it uses the value of the environment variable `AZURE_CLIENT_ID`.

Without this case handling, I could not inject a configuration without an ApiKey.

abestel: Hey @narengogi have you had time to check my comment above?

Here is a test (not sure I should commit it) that shows the kind of config that should work but that can't be validated without this check:

```typescript
import { configSchema } from './config';

describe('config schema refinement', () => {
  it('should validate config with azure-openai without api-key if auth mode managed is provided', () => {
    const workloadIdentityTest = {
      provider: 'azure-openai',
      azure_auth_mode: 'managed',
      resource_name: 'oai-abestel',
      deployment_id: 'gpt-4o',
      api_version: '2024-08-01-preview',
    };

    let result = configSchema.safeParse(workloadIdentityTest);
    expect(result.success).toBeTruthy();
  });

  it('should validate config with azure-openai without api-key if auth mode workload is provided', () => {
    const workloadIdentityTest = {
      provider: 'azure-openai',
      azure_auth_mode: 'workload',
      resource_name: 'oai-abestel',
      deployment_id: 'gpt-4o',
      api_version: '2024-08-01-preview',
    };

    let result = configSchema.safeParse(workloadIdentityTest);
    expect(result.success).toBeTruthy();
  });
});
```

narengogi (approved):

src/middlewares/requestValidator/schema/config.ts
narengogi: lgtm, thank you for the clarification @abestel, much appreciated

src/providers/azure-openai/api.ts
```diff
@@ -39,6 +41,40 @@ const AzureOpenAIAPIConfig: ProviderAPIConfig = {
...
+        tenantId &&
+        clientId &&
+        federatedTokenFile &&
+        runtime === 'node'
+      ) {
+        const fs = await import('fs');
+        const federatedToken = fs.readFileSync(federatedTokenFile, 'utf8');
+
+        if (federatedToken) {
+          const scope = 'https://cognitiveservices.azure.com/.default';
```
b4s36t4, line 64: We can remove the `scope` from here and just handling inside the `getAzureWorkloadIdentityToken`.

src/providers/azure-openai/api.ts
```diff
@@ -39,6 +41,40 @@ const AzureOpenAIAPIConfig: ProviderAPIConfig = {
...
+      const runtime = getRuntimeKey();
+      if (
+        authorityHost &&
+        tenantId &&
+        clientId &&
+        federatedTokenFile &&
+        runtime === 'node'
+      ) {
+        const fs = await import('fs');
+        const federatedToken = fs.readFileSync(federatedTokenFile, 'utf8');
```
b4s36t4, line 61: @VisargD Please share your opinion here, we can also accept the `contents` of `AZURE_FEDERATED_TOKEN_FILE ` file from header and parse it here. With this we can also make sure this works with cloudflare env as well.

b4s36t4 (changes requested):

b4s36t4:
hey, @abestel Sorry for back and forth changes, everything looks good to me, just a minor change. Will merge after that.

b4s36t4:
Hey, @abestel. Due to longer inactivity we're closing this PR. You changes will be part of #1408 PR which includes your commits as well. Thanks a lot for your contribution!.

PR #947, opened 2025-02-18
Added request json as params to conditional routing

roh26it:
Would open up model based conditional routing

### DIFF
--- a/src/handlers/handlerUtils.ts
+++ b/src/handlers/handlerUtils.ts
@@ -750,10 +750,21 @@ export async function tryTargetsRecursively(
       } catch (err) {
         metadata = {};
       }
+
+      let params =
+        request instanceof FormData ||
+        request instanceof ReadableStream ||
+        request instanceof ArrayBuffer
+          ? {} // Send empty object if not JSON
+          : request;
+
       let conditionalRouter: ConditionalRouter;
       let finalTarget: Targets;
       try {
-        conditionalRouter = new ConditionalRouter(currentTarget, { metadata });
+        conditionalRouter = new ConditionalRouter(currentTarget, {
+          metadata,
+          params,
+        });
         finalTarget = conditionalRouter.resolveTarget();
       } catch (conditionalRouter: any) {
         throw new RouterError(conditionalRouter.message);
--- a/src/services/conditionalRouter.ts
+++ b/src/services/conditionalRouter.ts
@@ -6,6 +6,7 @@ type Query = {
 
 interface RouterContext {
   metadata?: Record<string, string>;
+  params?: Record<string, any>;
 }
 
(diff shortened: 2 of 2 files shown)


src/handlers/handlerUtils.ts
```diff
@@ -750,10 +750,21 @@ export async function tryTargetsRecursively(
...
+        request instanceof FormData ||
+        request instanceof ReadableStream ||
+        request instanceof ArrayBuffer
+          ? {} // Send empty object if not JSON
+          : request;
+
       let conditionalRouter: ConditionalRouter;
       let finalTarget: Targets;
       try {
-        conditionalRouter = new ConditionalRouter(currentTarget, { metadata });
```
narengogi, line 769: shouldn't this be backward compatible with routing using metadata as well? are you removing support for it?

src/handlers/handlerUtils.ts
```diff
@@ -750,10 +750,21 @@ export async function tryTargetsRecursively(
       } catch (err) {
         metadata = {};
       }
+
+      let params =
```
narengogi, line 767: this is a good check, but I would recommend extracting the conditional routing code snippet and return early if the body is not of the `Record<string, string>` type, that would be faster.

```js
const getTargetFromConditionalRouter = () => {}
```

narengogi:
@roh26it also please setup a signing key and add it to GitHub, we have a policy that requires signed commits 

https://docs.github.com/en/authentication/managing-commit-signature-verification

once you setup your key, you'd need to remove these commits and create a new commit, alternatively you could use another branch and cherrypick commits from here

to reset and commit again, so you can do
```sh
git reset --soft HEAD~n      # n is number of commits
git commit -m "signed commit"
git push --force             # force is required as you're removing commits from upstream
```

src/handlers/handlerUtils.ts
VisargD: Hey, metadata is still present. This PR just adds one more type of context property named `params`

src/handlers/handlerUtils.ts
roh26it: Returning might not be an option since this is just adding context to the router. We don't manipulate any keys within the context in there. If anything is present, only then it gets picked up.

Are you saying that the type checks would be slow here or the 3 if statements?

VisargD (approved):

src/handlers/handlerUtils.ts
VisargD: Resolving this

src/handlers/handlerUtils.ts
VisargD: Current approach looks good to me. We can revisit this if required

Issue #722, opened 2024-11-01
Gateway error when processing non-200 model response


mkrueger12:
## Issue Description
The gateway fails to properly parse error messages when accessing Claude models on GCP Vertex AI via the streaming endpoint. The issue occurs when the model returns a non-200 response.

## Environment
- Gateway Endpoint: `/v1/chat/completions`
- Model: `anthropic.claude-3-5-sonnet@20240620`
- Provider: GCP Vertex AI
- Gateway is locally hosted

## Configuration
```json
{
  "strategy": {
    "mode": "loadbalance"
  },
  "targets": [
    {
      "provider": "vertex-ai",
      "vertex_project_id": "dev",
      "vertex_service_account_json": "SA",
      "vertex_region": "us-east5",
      "weight": 1
    }
  ]
}
```

## Error Output
Gateway logs:
```
event: error
2024-11-01 11:43:41 ^
2024-11-01 11:43:41 
2024-11-01 11:43:41 SyntaxError: Unexpected token 'e', "event: err"... is not valid JSON
2024-11-01 11:43:41     at JSON.parse (<anonymous>)
2024-11-01 11:43:41     at Xt (file:///app/build/start-server.js:2:71350)
2024-11-01 11:43:41     at file:///app/build/start-server.js:2:146756
2024-11-01 11:43:41     at async file:///app/build/start-server.js:2:146361
```

Application logs:
```python
requests.exceptions.ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
```

## Root Cause Analysis
- The error occurs when the model returns a non-200 response
- Verified by running parallel requests directly against the GCP API

## Additional Notes
- This issue reproduces with other configurations, including fallback mode
- The error is not specific to the provided configuration

### What Should Have Happened?

The gateway should return an appropriate error response to the application.

Hitting the GCP api directly returns:
```json
529 {"type":"error","error":{"type":"overloaded_error","message":"Overloaded"}}
```

### Your Twitter/LinkedIn

https://www.linkedin.com/in/maxkrueger1/

mkrueger12:
```
2024-11-02 14:21:05 Your AI Gateway is now running on http://localhost:8787 🚀
2024-11-02 14:21:07 chunk event: error
2024-11-02 14:21:07 data: {"type":"error","error":{"details":null,"type":"overloaded_error","message":"Overloaded"}           }
2024-11-02 14:21:07 undefined:1
[...]

PR #949, opened 2025-02-21
fix: update logic for parsing errors while streaming in anthropic

narengogi:
https://github.com/Portkey-AI/gateway/issues/722

we've previously added changes for this here, but missed removing the chunk event type header, hence the error was never caught
https://github.com/Portkey-AI/gateway/pull/741

### DIFF
--- a/src/providers/anthropic/chatComplete.ts
+++ b/src/providers/anthropic/chatComplete.ts
@@ -485,6 +485,7 @@ export const AnthropicChatCompleteStreamChunkTransform: (
   chunk = chunk.replace(/^event: content_block_start[\r\n]*/, '');
   chunk = chunk.replace(/^event: message_delta[\r\n]*/, '');
   chunk = chunk.replace(/^event: message_start[\r\n]*/, '');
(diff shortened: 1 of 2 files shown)


src/providers/anthropic/chatComplete.ts
```diff
@@ -485,6 +485,7 @@ export const AnthropicChatCompleteStreamChunkTransform: (
   chunk = chunk.replace(/^event: content_block_start[\r\n]*/, '');
   chunk = chunk.replace(/^event: message_delta[\r\n]*/, '');
   chunk = chunk.replace(/^event: message_start[\r\n]*/, '');
+  chunk = chunk.replace(/^event: error/, '');
```
VisargD: Hey, can you please follow the same regex pattern as other events. We will need `[\r\n]*`

narengogi: doing

narengogi: updated

VisargD (approved):

Issue #953, opened 2025-02-21
[Feature] Support the new openai routes as unified routes


narengogi:
### What Would You Like to See with the Gateway?

openai has added new routes:

listChatCompletions
getChatCompletion
getChatCompletionMessages
updateChatCompletion (for updating metadata)
deleteChatCompletion

refer: https://platform.openai.com/docs/api-reference/chat

PR #954, opened 2025-02-21
feat: support new openai routes as unified endpoints

narengogi:
#953 

This adds support for the new chat completions endpoints for both openai and azure openai
https://platform.openai.com/docs/api-reference/chat
https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/stored-completions?tabs=python-secure

1. GET /chat/completions/:completionId
2. GET /chat/completions
3. POST /chat/completions/:completionId
4. GET /chat/completions/:completionId/messages
5. DELETE /chat/completions/:completionId

**Testing Done**
- [X] Tested all the new routes for openai
- [ ] Tested all the new routes for azure-openai

### DIFF
--- a/src/providers/azure-openai/api.ts
+++ b/src/providers/azure-openai/api.ts
@@ -76,13 +76,6 @@ const AzureOpenAIAPIConfig: ProviderAPIConfig = {
       }
     }
 
-    const segments = gatewayRequestURL?.split('/');
-    let id = '';
-    if (['retrieveFileContent'].includes(mappedFn)) {
-      id = segments?.at(-2) ?? '';
-    } else {
-      id = segments?.at(-1) ?? '';
-    }
     const path = gatewayRequestURL.split('/v1')?.[1];
     const apiVersionQueryParam = path.includes('?')
       ? `&api-version=${apiVersion}`
--- a/src/providers/azure-openai/api.ts
+++ b/src/providers/azure-openai/api.ts
@@ -84,7 +84,9 @@ const AzureOpenAIAPIConfig: ProviderAPIConfig = {
       id = segments?.at(-1) ?? '';
     }
     const path = gatewayRequestURL.split('/v1')?.[1];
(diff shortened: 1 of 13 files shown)


src/providers/azure-openai/api.ts
```diff
@@ -77,6 +77,9 @@ const AzureOpenAIAPIConfig: ProviderAPIConfig = {
     }
 
     const path = gatewayRequestURL.split('/v1')?.[1];
+    const apiVersionQueryParam = path.includes('?')
+      ? `&api-version=${apiVersion}`
+      : `?api-version=${apiVersion}`;
```
b4s36t4: This is a bit confusing logic and can lead to unwanted issues and maintenance. instead we can use the following logic?
```js
url = new URL(gatewayURL)
url.searchParams.set('api-version',apiVersion)

// using the params.

`/deployments/${deploymentId}/${path}?${url.searchParams.toString()}`
```

This way we can be sure that we don't mess-up the queryparams in any edge case.

src/providers/openai/updateChatCompletions.ts
```diff
@@ -0,0 +1,25 @@
...
+import { ErrorResponse, ProviderConfig } from '../types';
+import { OpenAIChatCompleteResponse } from './chatComplete';
+import { OpenAIErrorResponseTransform } from './utils';
+
+export const OpenAIUpdateChatCompletionConfig: ProviderConfig = {
+  metadata: {
+    param: 'metadata',
+    required: true,
+  },
+};
```
b4s36t4, line 11: I'd recommend use the same config for `azure-openai` as well, both of the APIs share same signature.

b4s36t4 (changes requested):

src/providers/azure-openai/api.ts
narengogi: thanks, that's much cleaner but the path variable here also has the existing query params, I'll separate them

VisargD:
Hi @narengogi , closing this PR as it has been open for a very long time and it seems to be outdated now. Feel free to reopen the PR once the changes are ready. Thanks!

Issue #964, opened 2025-02-26
[Feature] Return provider raw/untransformed response when 1:1 mapping with openai fields is not possible


narengogi:
### What Would You Like to See with the Gateway?

Portkey gateway is openai compliant, but of late there are cases in which other providers like anthropic and vertex have API signatures that do not have 1:1 mapping with openai fields.

While there has been no decision on moving from the openai signature to a unified signature so far (most of the new anthropic and vertex features are in beta), there is still a need for the gateway to support the new features.

For now, the proposed solution is to return the raw response along with the transformed response in these cases with a new header `x-portkey-include-raw-response: true`

### Context for your Request

Take for example the case of [anthropic thinking mode](https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking#streaming-extended-thinking)

The returned content object is an array with multiple text and thinking items like so:
```json
{
  "content": [
    {
      "type": "thinking",
      "thinking": "Let me analyze this step by step...",
      "signature": "WaUjzkypQ2mUEVM36O2TxuC06KN8xyfbJwyem2dw3URve/op91XWHOEBLLqIOMfFG/UvLEczmEsUjavL...."
    },
    {
      "type": "redacted_thinking",
      "data": "EmwKAhgBEgy3va3pzix/LafPsn4aDFIT2Xlxh0L5L8rLVyIwxtE3rAFBa8cr3qpP..."
    },
    {
      "type": "text",
      "text": "Based on my analysis..."
    }
  ]
}
```

openai's `content` field only supports content as a string.

3 approaches considered:

1. return the content as an array when `x-portkey-strict-open-ai-compliance: false` header is sent
this is a good option, but it'll break the SDK support as content would be expected to be a string, and the SDK sends `x-portkey-strict-open-ai-compliance: false` by default and the flag was not meant to be used for this
2. serialize the content array and send it as a string when `x-portkey-strict-open-ai-compliance: false` is set. this is a pretty good option, but it's not of much use when consuming streams, and newer features like this keep coming up
3. allow users to consume raw response when there is no one to one mapping with a new raw_response field

narengogi:
closing this as won't do as the gateway returns a unified signature now

PR #965, opened 2025-02-26
support returning raw_response for anthropic for enabling thinking mode as a beta feature

narengogi:
**Related Issues:**
- #964 

To receive `raw_response` along with the transformed response, please include the following header when making your requests: `x-portkey-include-raw-response: true` 

**Testing Done**
- [X] non-streaming anthropic calls with `x-portkey-include-raw-response` set to false and true
- [X] streaming anthropic calls with `x-portkey-include-raw-response` set to false and true
- [X] Test one other provider with streaming and non-streaming to make sure the streamHandler changes are working correctly

**Note**
this PR adds support for raw_response in non-streaming responses only

### DIFF
--- a/src/handlers/streamHandler.ts
+++ b/src/handlers/streamHandler.ts
@@ -13,6 +13,28 @@ import { OpenAIChatCompleteResponse } from '../providers/openai/chatComplete';
 import { OpenAICompleteResponse } from '../providers/openai/complete';
 import { getStreamModeSplitPattern, type SplitPatternType } from '../utils';
 
+const appendRawResponseToChunks = (
+  chunk: string,
+  rawResponse: Record<string, any>
+) => {
+  try {
+    if (chunk !== undefined) {
+      const parts = chunk.split('\n\n');
+      for (let i = 0; i < parts.length; i++) {
+        let part = parts[i];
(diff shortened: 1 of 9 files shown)


src/handlers/streamHandler.ts
```diff
@@ -13,6 +13,28 @@ import { OpenAIChatCompleteResponse } from '../providers/openai/chatComplete';
...
+const appendRawResponseToChunks = (
+  chunk: string,
+  rawResponse: Record<string, any>
+) => {
+  try {
+    if (chunk !== undefined) {
+      const parts = chunk.split('\n\n');
+      for (let i = 0; i < parts.length; i++) {
+        let part = parts[i];
+        if (part.startsWith('data: ')) {
```
VisargD: Hey, wouldn't this break for providers like anthropic which sends event tags prefix along with `data:` in stream chunks.

narengogi: hey, the chunks here are transformed chunks in openai format, so they won't have any event tags

narengogi:
closing this in favor of a unified API approach as discussed with @VisargD and @ayush-portkey 
The unified spec will follow anthropics API signature as it is similar to openai signature, will add details in a github issue

Issue #893, opened 2025-01-27
Error messages due to invalid logging optimization


em-al-wi:
### What Happened?

Since version 1.8.4 error messages in the log appear whenever a service response is not a JSON document or the JSON is longer then 1,000 characters.

`ERROR SyntaxError: Unterminated string in JSON at position 1003
     at JSON.parse (<anonymous>)
     at Fi (file:///usr/src/app/node_modules/.pnpm/@portkey-ai+gateway@1.9.3/node_modules/@portkey-ai/gateway/build/start-server.js:2:260672)
     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)`

### What Should Have Happened?

No error messages.

### Relevant Code Snippet

The commit introducing the error: https://github.com/Portkey-AI/gateway/commit/cf25af712dff25a260c6426fc9617445f2f42d5c

The change does not take into account that a response is not always json. And if it is json but longer than 1,000 characters, it truncates the json and tries to parse it.

**src/middlewares/log/index.ts**:
`JSON.parse(responseString.substring(0, maxLength) + '...') `

narengogi:
@em-al-wi  got it, will remove the check, 
do you want to take this up?

em-al-wi:
@narengogi It would be greatly appreciated if you could remove the check. This also becomes messy when using guardrails, as the response is usually more than 1,000 characters in length then.

urbanonymous:
Happening here as well. Also getting Unterminated fractional number in JSON at position 1001.

portkey v1.8.3 (can't update because other bug that you have breaks the portkey server)

urbanonymous:
#970 should fix this

narengogi:
big thanks to @urbanonymous for the contribution, closing this as completed now

PR #970, opened 2025-03-03
fix: logging error with more than 1k characters

urbanonymous:
**Fix logging error with more than 1k characters:** 
- In this PR the 1000 characters limit from the 

**Description:**
- Extract the magic number to a constant
- Set the constant into 100k

**Motivation:** 
- Tons of SyntaxtError logs in the console

**Related Issues:**
- #893 

The simplest and most effective fix is to set MAX_RESPONSE_LENGTH to a higher, practical value that accommodates typical LLM outputs while keeping log messages manageable for broadcasting to clients via WebSockets. However, we must also modify the code to prevent parsing errors when truncation occurs. Here’s why and how:

- Why Not Remove the Check? While removing the length check would log full responses, it risks sending extremely large messages (e.g., megabytes) to clients, potentially causing performance issues in WebSocket communication or client handling. A reasonable limit provides balance.

- Why a Higher Limit? LLM outputs can easily exceed 1000 characters. For example, a 4096-token response (a common max for models like GPT-3) might translate to ~16,000 characters or more in JSON form, including metadata. A higher limit ensures most responses are logged fully.

- Why Fix Truncation? Even with a higher limit, responses exceeding it must be handled gracefully to avoid invalid JSON errors.

## Setting MAX_RESPONSE_LENGTH to 100,000 characters (~100KB):

- Covers Most Cases: This captures full responses for most LLM outputs, including lengthy ones up to tens of thousands of characters, while accounting for JSON overhead (e.g., keys, quotes).

- WebSocket Friendly: Modern WebSocket implementations, including in Cloudflare Workers (where your app runs with workerd), support messages up to 1MB. At ~100KB, this is well within limits and safe for broadcasting.

- Practical for Logging: It balances completeness with preventing excessively large log entries.

### DIFF
--- a/src/middlewares/log/index.ts
+++ b/src/middlewares/log/index.ts
@@ -68,9 +68,14 @@ async function processLog(c: Context, start: number) {
       : await c.res.clone().json();
 
     const responseString = JSON.stringify(response);
-    requestOptionsArray[0].response = responseString.length > MAX_RESPONSE_LENGTH
-      ? JSON.parse(responseString.substring(0, MAX_RESPONSE_LENGTH) + '...')
-      : response;
+    if (responseString.length > MAX_RESPONSE_LENGTH) {
+      requestOptionsArray[0].response = {
+        ...response,
+        __truncated: `Response truncated (${responseString.length} chars exceeded limit of ${MAX_RESPONSE_LENGTH})`
+      };
+    } else {
+      requestOptionsArray[0].response = response;
+    }
   } catch (error) {
     console.error('Error processing log:', error);
   }
--- a/src/middlewares/log/index.ts
+++ b/src/middlewares/log/index.ts
@@ -69,7 +69,8 @@ async function processLog(c: Context, start: number) {
 
     const responseString = JSON.stringify(response);
     if (responseString.length > MAX_RESPONSE_LENGTH) {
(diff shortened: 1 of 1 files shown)


src/middlewares/log/index.ts
```diff
@@ -61,18 +62,17 @@ async function processLog(c: Context, start: number) {
...
+      ? { message: 'The response was a stream.' }
+      : await c.res.clone().json();
+
     const responseString = JSON.stringify(response);
-    requestOptionsArray[0].response =
-      responseString.length > maxLength
-        ? JSON.parse(responseString.substring(0, maxLength) + '...')
-        : response;
+    requestOptionsArray[0].response = responseString.length > MAX_RESPONSE_LENGTH
+      ? JSON.parse(responseString.substring(0, MAX_RESPONSE_LENGTH) + '...')
```
narengogi: I think we can just return the substring instead instead of trying to deserialize the string

narengogi (dismissed):
lgtm

src/middlewares/log/index.ts
urbanonymous: fixed

VisargD (dismissed):

VisargD:
Hey @urbanonymous , PR looks good to go. Can you please run `npm run format` once?

VisargD (approved):

PR #983, opened 2025-03-11
Providers/ncompass

vineye25:
**Title:** 
nCompass Technologies Integration

**Description:** (optional)
Configured nCompass Technologies as a LLM provider


src/providers/ncompass/chatComplete.ts
```diff
@@ -0,0 +1,204 @@
...
+  ChatCompletionResponse,
+  ErrorResponse,
+  ProviderConfig,
+} from '../types';
+import {
+  generateErrorResponse,
+  generateInvalidProviderResponseError,
+} from '../utils';
+
+// TODOS: this configuration does not enforce the maximum token limit for the input parameter. If you want to enforce this, you might need to add a custom validation function or a max property to the ParameterConfig interface, and then use it in the input configuration. However, this might be complex because the token count is not a simple length check, but depends on the specific tokenization method used by the model.
```
narengogi: this is fine, provider will handle this validation and throw the relevant error, no point in maintaining a model specific mapping the code

src/providers/ncompass/chatComplete.ts
```diff
@@ -0,0 +1,204 @@
...
+    param: 'stop',
+    default: null,
+  },
+  stream: {
+    param: 'stream',
+    default: false,
+  },
+};
+
+interface NCompassChatCompleteResponse extends ChatCompletionResponse {
```
narengogi: all of the keys `id, object, created, mode, usage` are already present in ChatCompletionResponse, this is redundant

narengogi (commented):
Looks good to me, only a minor comment @vineye25

src/providers/ncompass/chatComplete.ts
vineye25: Sounds good, I've removed this "todo" in the latest commit

src/providers/ncompass/chatComplete.ts
vineye25: Makes sense, I had used `deepinfra/chatComplete.ts` as an example to work from but hadn't checked the base implementation for `ChatCompletionResponse`.

I've updated this in the latest commit

narengogi (approved):

narengogi (dismissed):

VisargD:
Hey @vineye25 , can you please run `npm run format`, the check workflow is failing.

PR #984, opened 2025-03-12
Add labels parameter in Vertex AI

elentaure:
**Title:** 
- Add labels parameter to the vertex provider config

**Description:** (optional)
- Added the labels parameter so custom metadata labels can be added to the call. Documentation about this feature: https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/add-labels-to-api-calls

### DIFF
--- a/src/providers/google-vertex-ai/chatComplete.ts
+++ b/src/providers/google-vertex-ai/chatComplete.ts
@@ -334,7 +334,6 @@ export const VertexGoogleChatCompleteConfig: ProviderConfig = {
   },
   labels: {
     param: 'labels',
-    default: '',
   },
 };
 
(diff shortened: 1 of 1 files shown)


src/providers/google-vertex-ai/chatComplete.ts
```diff
@@ -332,6 +332,10 @@ export const VertexGoogleChatCompleteConfig: ProviderConfig = {
       }
     },
   },
+  labels: {
+    param: 'labels',
+    default: '',
```
b4s36t4: Is this a required field for vertex? If not we can remove default property else please add `required` property with true to keep the default value in effect.

b4s36t4 (changes requested):

src/providers/google-vertex-ai/chatComplete.ts
elentaure: Its not required, so should I remove default then if not required?

b4s36t4: Yea, please.

elentaure: Removed

VisargD (approved):

PR #993, opened 2025-03-20
fix: remove conditional for bedrock list/get endpoints, date fixes

b4s36t4:
**Title:** 
- Bedrock list/get batch endpoints are not required `model`, we had a hard check which is not needed. Also some unwanted properties are handled with conditionals.

### DIFF
--- a/src/providers/bedrock/api.ts
+++ b/src/providers/bedrock/api.ts
@@ -1,7 +1,5 @@
-import { env } from 'hono/adapter';
 import { Context } from 'hono';
 import { Options } from '../../types/requestBody';
-import { GatewayError } from '../../errors/GatewayError';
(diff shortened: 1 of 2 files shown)


src/providers/bedrock/api.ts
```diff
@@ -169,9 +160,6 @@ const BedrockAPIConfig: BedrockAPIConfigInterface = {
       return `/model-invocation-job/${batchId}/stop`;
     }
     const { model, stream } = gatewayRequestBody;
-    if (!model && !BEDROCK_FINETUNE_ENDPOINTS.includes(fn as endpointStrings)) {
```
VisargD, line 196: Can we please keep atleast the model check

VisargD (dismissed):

PR #1033, opened 2025-04-11
fix: correctly handle other errors from handlers

jpulec:
Code Quality bug fix 

## Author Description

**Title:** 
This fixes an issue where some unhandled errors thrown by `tryPost` cause big issues. Due to some loose typing, it was possible for the handlers to return `null` for a `response`. This causes *bad things* as eventually it makes its way to the top of the call chain, and if you return from your last handler in hono, the context itself is returned, which is difficult to understand what is going on.

**Title:** fix: correctly handle other errors from handlers

### 🔄 What Changed
- Added explicit typing for `response` variable as `Response | null`
- Changed `error: any` to `error: unknown` for better type safety
- Added proper type checking for error objects before accessing properties
- Added a safeguard that throws an error when no response is created

### 🔍 Impact of the Change
- Prevents unhandled errors from causing unexpected behavior
- Ensures that a proper response is always returned or an error is thrown
- Improves type safety throughout the error handling flow
- Provides clearer error messages when something goes wrong

### 📁 Total Files Changed
- 1 file modified: `src/handlers/handlerUtils.ts` (12 additions, 3 deletions)

### 🧪 Test Added
N/A

N/A

## Quality Recommendations

1. Consider adding unit tests to verify the error handling behavior

2. Add more specific error messages that include details about what went wrong

3. Consider implementing a more robust error handling strategy with custom error types

### DIFF
--- a/src/handlers/handlerUtils.ts
+++ b/src/handlers/handlerUtils.ts
@@ -737,7 +737,7 @@ export async function tryTargetsRecursively(
   ];
   // end: merge inherited config with current target config (preference given to current)
 
-  let response;
+  let response: Response | null = null;
 
   switch (strategyMode) {
     case StrategyModes.FALLBACK:
@@ -859,7 +859,7 @@ export async function tryTargetsRecursively(
           currentJsonPath,
           method
         );
-      } catch (error: any) {
+      } catch (error: unknown) {
         // tryPost always returns a Response.
         // TypeError will check for all unhandled exceptions.
         // GatewayError will check for all handled exceptions which cannot allow the request to proceed.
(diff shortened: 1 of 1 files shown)


src/handlers/handlerUtils.ts
```diff
@@ -859,7 +859,7 @@ export async function tryTargetsRecursively(
           currentJsonPath,
           method
         );
-      } catch (error: any) {
+      } catch (error: unknown) {
```
narengogi, line 862: I've just gone through this code, and I think this catch block is meant to set a response object in all cases.
so the correct fix to make sure all cases are handled would be
```js
if (typeof error === 'object' &&
      error &&
          'response' in error &&
          error.response instanceof Response
        ) {
          response = error.response;
        } else if (error instanceof GatewayError) {...} else {response = new Response(500,...)}
```

narengogi: @VisargD to take a call here

VisargD:
Hi! We have made some major refactoring in the core methods and the issue that you have mentioned is already handled in that. Please let us know if you still see any issues in the latest main branch methods for this. Thanks!

PR #1067, opened 2025-04-26
feat: transform raw image responses to json

unsync:
Code Quality new feature 

### 🔄 What Changed
This PR adds functionality to transform image content type responses from Segmind to JSON format. It introduces a new handler for converting image responses to base64-encoded JSON, specifically targeting the imageGenerate route. The implementation includes error handling to fall back to the original response if transformation fails.

### 🔍 Impact of the Change
This change improves the handling of image responses by standardizing them to JSON format, making it easier for clients to process the responses. It's designed as a non-breaking change that enhances the existing functionality without disrupting current behavior.

### 📁 Total Files Changed
5 files changed with 67 additions and 1 deletion:
- Added new file: src/handlers/imageToJsonHandler.ts
- Modified: src/handlers/responseHandlers.ts
- Modified: src/handlers/retryHandler.ts
- Modified: src/providers/segmind/imageGenerate.ts
- Modified: src/providers/segmind/index.ts

### 🧪 Test Added
No tests have been added in this PR.

No security vulnerabilities detected.

## Type of Change
- [ ] Documentation update

## How Has This Been Tested?
- [ ] Unit Tests
- [ ] Integration Tests
- [x] Manual Testing

## Screenshots (if applicable)
N/A

## Checklist

## Related Issues
N/A

## Quality Recommendations

1. Add unit tests for the new image transformation functionality

2. Add more detailed error logging in imageToJsonHandler.ts with specific error types

3. Consider adding a content-type validation function to avoid repetitive checks

4. Add documentation for the new image transformation feature

## Sequence Diagram

```mermaid
sequenceDiagram
    participant Client
    participant ResponseHandler
    participant RetryHandler
    participant ImageToJsonHandler
    participant SegmindProvider

    Client->>ResponseHandler: Request image generation
    ResponseHandler->>RetryHandler: Process request with retry logic
    RetryHandler->>SegmindProvider: Call Segmind API
    SegmindProvider-->>RetryHandler: Return image response
    RetryHandler-->>ResponseHandler: Return response
    
    Note over ResponseHandler: Check if response needs transformation
    alt responseTransformer === 'imageGenerate' && has imageToJson transformer
        ResponseHandler->>ImageToJsonHandler: getTransformedResponse(response, transformer)
        Note over ImageToJsonHandler: Convert image to base64
        ImageToJsonHandler-->>ResponseHandler: Return JSON response
    end
    
    ResponseHandler-->>Client: Return final response
```

### DIFF
--- a/src/handlers/imageToJsonHandler.ts
+++ b/src/handlers/imageToJsonHandler.ts
@@ -1,0 +1,26 @@
+import { CONTENT_TYPES } from '../globals';
+
+export async function getTransformedResponse(_: {
+  response: Response;
+  transformer: (base64Image: string) => Record<string, any>;
+}): Promise<Response | undefined> {
+  const { response, transformer } = _;
+
+  console.info('imageToJsonHandler > converting image response to base64 JSON');
+  const imageBuffer = await response.arrayBuffer();
+  // Simple ArrayBuffer to base64 conversion for environments like Cloudflare Workers
+  let binary = '';
+  const bytes = new Uint8Array(imageBuffer);
+  const len = bytes.byteLength;
+  for (let i = 0; i < len; i++) {
+    binary += String.fromCharCode(bytes[i]);
+  }
+  return new Response(JSON.stringify(transformer(btoa(binary))), {
+    headers: {
+      ...Object.fromEntries(response.headers), // keep original headers
+      'content-type': CONTENT_TYPES.APPLICATION_JSON,
+    },
+    status: response.status,
+    statusText: response.statusText,
+  });
+}
--- a/src/handlers/imageToJsonHandler.ts
+++ b/src/handlers/imageToJsonHandler.ts
@@ -5,22 +5,31 @@ export async function getTransformedResponse(_: {
   transformer: (base64Image: string) => Record<string, any>;
 }): Promise<Response | undefined> {
   const { response, transformer } = _;
-
-  console.info('imageToJsonHandler > converting image response to base64 JSON');
-  const imageBuffer = await response.arrayBuffer();
-  // Simple ArrayBuffer to base64 conversion for environments like Cloudflare Workers
-  let binary = '';
-  const bytes = new Uint8Array(imageBuffer);
-  const len = bytes.byteLength;
-  for (let i = 0; i < len; i++) {
-    binary += String.fromCharCode(bytes[i]);
+  try {
+    console.info(
+      'imageToJsonHandler > converting image response to base64 JSON'
+    );
+    const imageBuffer = await response.arrayBuffer();
+    // Simple ArrayBuffer to base64 conversion for environments like Cloudflare Workers
+    let binary = '';
+    const bytes = new Uint8Array(imageBuffer);
+    const len = bytes.byteLength;
+    for (let i = 0; i < len; i++) {
+      binary += String.fromCharCode(bytes[i]);
+    }
+    return new Response(JSON.stringify(transformer(btoa(binary))), {
+      headers: {
+        ...Object.fromEntries(response.headers), // keep original headers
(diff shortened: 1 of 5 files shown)


narengogi (commented):
@unsync I have a few items in mind wrt this PR
1. we should be handling transforms only for specific routes and not based on response, so if segmind has a `/v1/images/generations` route, we can do these changes for that route
2. I'm not entirely keen we should load the entire image file into memory and convert it to json, but that's okay

overall I'm saying this PR could only include changes to `provider/segmind/imageGenerate.ts` and `provider/segmind/api.ts` and `provider/segmind/index.ts`

unsync:
@narengogi if i get this correctly, you would prefer to scope the changes to the providers handlers.

Unless i'm missing something, that would not work without a change in the reponseHandler because of this:

https://github.com/Portkey-AI/gateway/blob/de327f6886d542d3929ce63dd64710ba335fdbb2/src/handlers/responseHandlers.ts#L137-L139

This prevents to reach the code that handles the transformations:

https://github.com/Portkey-AI/gateway/blob/de327f6886d542d3929ce63dd64710ba335fdbb2/src/handlers/responseHandlers.ts#L159-L166

And before you hit `handleNonStreamingMode`, you need to have a JSON body, otherwise this fails:

https://github.com/Portkey-AI/gateway/blob/de327f6886d542d3929ce63dd64710ba335fdbb2/src/handlers/streamHandler.ts#L248-L250


Since the provider returns a content type `image/` then we don't enter the transformer function of the providers (if we did, i definitely would have gone this road, that was my first idea ^^).

Let me know if i missed something here 🙏 

And if that doesn't makes it's way to `main` no hard feelings, i'm using this on a fork, i just wanted for others to benefit from it 👍

narengogi:
I spent some time on this, I was initially misreading what the PR is doing, this fix makes sense for the problem, 

one change we could do is to convert to json only if routes are in the unified routes like `v1/images/generations`

I'm copying @VisargD on this for a review of the approach

unsync:
> I spent some time on this, I was initially misreading what the PR is doing, this fix makes sense for the problem,
> 
> one change we could do is to convert to json only if routes are in the unified routes like `v1/images/generations`
> 
> I'm copying @VisargD on this for a review of the approach

@narengogi @VisargD i made a last commit to the PR to, imho, limit the scope of the behaviour in a more elegant way:

```
  // if the original response is an image, convert it to JSON if the provider has a transformer
  if (
    responseTransformer === 'imageGenerate' && // check that we are on the imageGenerate route
    responseTransformerFunction && // check that we have a transformer for this provider
    providerTransformers?.[`imageToJson`] && // check that we have a 'imageToJson" transformer for this provider
    response.headers
      ?.get('content-type')
      ?.startsWith(CONTENT_TYPES.GENERIC_IMAGE_PATTERN) // check that the original response content type is an image
  ) {
    // transformers are async, because we read the body as an array buffer
    response = await providerTransformers?.[`imageToJson`](response);
  }
```


This ensures that : 
- we only apply it when the response is initially a `image/***`
- we are looking for a `responseTransformer` of value `imageGenerate` (ensuring to do this only on image generation)
- the provider has the `responseTransformerFunction` (no point transforming the response if the provider doesn't provide a transformer)
- obviously, checks if the provider has a `imageToJson` transformer

I also added a try/catch the the transformer wrapper, so that if the transformation fails we return the original response to avoid any failures due to this logic

VisargD:
@narengogi had some ideas for this PR. Will wait for him to add more comments

unsync:
Hey !
I didn't rebase the PR in a while, but let me know if i should do it in case this is still considered

narengogi:
closing as stale

PR #1068, opened 2025-04-27
fix: return tool calls in non streaming openrouter responses

unsync:
Code Quality bug fix 

## Author Description

This PR fixes an issue that causes tool_calls not to be returned when not streaming the response.

This is due to the too strict mapping of messages.
Ideally, we could just pass the whole message content, like it is done with the streaming responses.

### 🔄 What Changed
This PR fixes an issue where tool_calls were not being returned in non-streaming responses from OpenRouter. The implementation now properly includes tool_calls in the message object when transforming the response. Additionally, it adds support for content_blocks to handle reasoning tokens in a unified way across streaming and non-streaming responses.

### 🔍 Impact of the Change
This change ensures consistent behavior between streaming and non-streaming responses, allowing applications to properly receive and process tool_calls regardless of the streaming mode. It also improves the handling of reasoning content by using a standardized content_blocks approach.

### 📁 Total Files Changed
1 file changed with 73 additions and 13 deletions (86 total changes)

### 🧪 Test Added
N/A

No security vulnerabilities detected

## Type of Change
- [ ] Documentation update

## How Has This Been Tested?
- [ ] Unit Tests
- [ ] Integration Tests
- [x] Manual Testing

## Screenshots (if applicable)
N/A

## Checklist

## Related Issues
Issue with tool_calls not being returned in non-streaming OpenRouter responses

## Quality Recommendations

1. Add unit tests to verify the fix for tool_calls in non-streaming responses

2. Consider adding error handling for malformed responses from OpenRouter

3. Add comments explaining the purpose of content_blocks and their relationship to reasoning

## Sequence Diagram

```mermaid
sequenceDiagram
    participant Client
    participant Gateway
    participant OpenRouter
    
    Client->>Gateway: Send chat completion request
    Gateway->>OpenRouter: Forward request to OpenRouter API
    OpenRouter-->>Gateway: Return response with tool_calls
    
    alt Streaming Mode
        Gateway->>Gateway: OpenrouterChatCompleteStreamChunkTransform()
        Note over Gateway: Already handles tool_calls correctly
    else Non-Streaming Mode
        Gateway->>Gateway: OpenrouterChatCompleteResponseTransform()
        Note over Gateway: Now properly includes tool_calls in message
        Note over Gateway: Adds content_blocks when not in strictOpenAiCompliance mode
    end
    
    Gateway-->>Client: Return transformed response with tool_calls
```

### DIFF
--- a/src/providers/openrouter/chatComplete.ts
+++ b/src/providers/openrouter/chatComplete.ts
@@ -146,12 +146,14 @@ export const OpenrouterChatCompleteResponseTransform: (
         message: {
           role: c.message.role,
           content: c.message.content,
-          content_blocks: [
-            {
-              type: 'thinking',
-              thinking: c.message.reasoning,
-            },
-          ],
+          ...(c.message.reasoning && {
+            content_blocks: [
+              {
+                type: 'thinking',
+                thinking: c.message.reasoning,
+              },
+            ],
+          }),
           ...(c.message.tool_calls && {
             tool_calls: c.message.tool_calls,
           }),
--- a/src/providers/openrouter/chatComplete.ts
+++ b/src/providers/openrouter/chatComplete.ts
@@ -146,9 +146,12 @@ export const OpenrouterChatCompleteResponseTransform: (
         message: {
           role: c.message.role,
           content: c.message.content,
-          ...(c.message.reasoning && {
-            reasoning: c.message.reasoning,
-          }),
+          content_blocks: [
+            {
+              type: 'thinking',
+              thinking: c.message.reasoning,
+            },
+          ],
           ...(c.message.tool_calls && {
             tool_calls: c.message.tool_calls,
           }),
--- a/src/providers/openrouter/chatComplete.ts
+++ b/src/providers/openrouter/chatComplete.ts
@@ -1,6 +1,7 @@
 import { OPENROUTER } from '../../globals';
-import { Params } from '../../types/requestBody';
+import { Message, Params } from '../../types/requestBody';
 import {
+  ChatChoice,
   ChatCompletionResponse,
   ErrorResponse,
   ProviderConfig,
@@ -88,6 +89,7 @@ interface OpenrouterChatCompleteResponse extends ChatCompletionResponse {
     completion_tokens: number;
     total_tokens: number;
   };
+  choices: (ChatChoice & { message: Message & { reasoning: string } })[];
 }
 
 export interface OpenrouterErrorResponse {
@@ -112,6 +114,7 @@ interface OpenrouterStreamChunk {
     delta: {
       role?: string | null;
       content?: string;
+      reasoning?: string;
     };
     index: number;
     finish_reason: string | null;
@@ -120,8 +123,19 @@ interface OpenrouterStreamChunk {
 
 export const OpenrouterChatCompleteResponseTransform: (
   response: OpenrouterChatCompleteResponse | OpenrouterErrorResponse,
-  responseStatus: number
-) => ChatCompletionResponse | ErrorResponse = (response, responseStatus) => {
+  responseStatus: number,
+  _responseHeaders: Headers,
+  strictOpenAiCompliance: boolean,
+  _gatewayRequestUrl: string,
+  gatewayRequest: Params
+) => ChatCompletionResponse | ErrorResponse = (
+  response,
+  responseStatus,
+  _responseHeaders,
+  strictOpenAiCompliance,
+  _gatewayRequestUrl,
+  gatewayRequest
+) => {
   if ('message' in response && responseStatus !== 200) {
     return generateErrorResponse(
       {
@@ -146,14 +160,16 @@ export const OpenrouterChatCompleteResponseTransform: (
         message: {
           role: c.message.role,
           content: c.message.content,
-          ...(c.message.reasoning && {
-            content_blocks: [
-              {
-                type: 'thinking',
-                thinking: c.message.reasoning,
-              },
-            ],
-          }),
+          ...(!strictOpenAiCompliance &&
+            c.message.reasoning && {
+              // this is to match portkey unified api
+              content_blocks: [
+                {
+                  type: 'thinking',
+                  thinking: c.message.reasoning,
+                },
+              ],
+            }),
           ...(c.message.tool_calls && {
             tool_calls: c.message.tool_calls,
           }),
@@ -181,7 +197,7 @@ export const OpenrouterChatCompleteStreamChunkTransform: (
   responseChunk,
   fallbackId,
   _streamState,
-  _strictOpenAiCompliance,
+  strictOpenAiCompliance,
   gatewayRequest
 ) => {
   let chunk = responseChunk.trim();
@@ -216,7 +232,21 @@ export const OpenrouterChatCompleteStreamChunkTransform: (
       choices: [
         {
           index: parsedChunk.choices?.[0]?.index,
-          delta: parsedChunk.choices?.[0]?.delta,
+          delta: {
+            ...parsedChunk.choices?.[0]?.delta,
+            ...(!strictOpenAiCompliance &&
+              parsedChunk.choices?.[0]?.delta?.reasoning && {
+                // this is to match portkey unified api
+                content_blocks: [
+                  {
+                    index: parsedChunk.choices?.[0]?.index,
+                    delta: {
+                      thinking: parsedChunk.choices?.[0]?.delta?.reasoning,
+                    },
+                  },
+                ],
+              }),
+          },
           finish_reason: parsedChunk.choices?.[0]?.finish_reason,
         },
       ],
--- a/src/providers/openrouter/chatComplete.ts
+++ b/src/providers/openrouter/chatComplete.ts
@@ -146,6 +146,12 @@ export const OpenrouterChatCompleteResponseTransform: (
         message: {
           role: c.message.role,
           content: c.message.content,
+          ...(c.message.reasoning && {
+            reasoning: c.message.reasoning,
+          }),
+          ...(c.message.tool_calls && {
+            tool_calls: c.message.tool_calls,
+          }),
         },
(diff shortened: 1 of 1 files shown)


src/providers/openrouter/chatComplete.ts
```diff
@@ -146,6 +146,12 @@ export const OpenrouterChatCompleteResponseTransform: (
         message: {
           role: c.message.role,
           content: c.message.content,
+          ...(c.message.reasoning && {
```
narengogi: Hey @unsync 
the portkey api spec supports retuning reasoning messages in the response in the anthropic format, inside the key called `content_blocks`, we shouldn't be sending it at the top level since we have a unified signature now.

checkout the documentation here https://portkey.ai/docs/api-reference/inference-api/chat#thinking-content-part
image

narengogi (changes requested):

src/providers/openrouter/chatComplete.ts
unsync: @narengogi thanks for the feedback, i wasn't aware of this since the last time i worker on providers.

Could it be ok to add it anyway behind the `strictOpenAiCompliance` flag ?
The idea is to be able to use it with (example) the openrouter vercel sdk, that expects the reasoning to be in this key.

unsync: @narengogi done in https://github.com/Portkey-AI/gateway/pull/1068/commits/1f3243de43dbc43ef750fdc4969c3591667f7947

unsync: and a quick fix here : https://github.com/Portkey-AI/gateway/pull/1068/commits/5cd47c0ee76b80c663c965c9815e06ef5f04b213

unsync: @narengogi i made changes to use the `content_blocks` in the last commit:

```
{
  "id": "gen-1745919064-AJHWFf9YPCc8KEQTcwwa",
  "object": "chat.completion",
  "created": 1745919064,
  "model": "anthropic/claude-3.7-sonnet:thinking",
  "provider": "openrouter",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "2 + 2 = 4",
        "content_blocks": [
          {
            "type": "thinking",
            "thinking": "This is a simple arithmetic question. 2 + 2 = 4."
          }
        ]
      },
      "finish_reason": "stop"
    }
  ],
  "usage": {
    "prompt_tokens": 45,
    "completion_tokens": 41,
    "total_tokens": 86
  }
}
```

### streamed response chunk
```
{
  "id": "gen-1745919147-ZQxoU4OhGp6MHAxxvMD9",
  "object": "chat.completion.chunk",
  "created": 1745919147,
  "model": "anthropic/claude-3.7-sonnet:thinking",
  "provider": "openrouter",
  "choices": [
    {
      "index": 0,
      "delta": {
        "role": "assistant",
        "content": "",
        "content_blocks": [
          {
            "index": 0,
            "delta": {
              "thinking": " is a very simple arithmetic question asking what 2 + "
            }
          }
        ]
      },
      "finish_reason": null
    }
  ]
}
```

narengogi: hey @unsync, sorry for the delayed response, but `thinking_blocks` should contain both the thinking and the regular message parts, so you might want to handle it like this, also we should only return a non openai compliant field when `strictOpenAiCompliance` is `false`
```js
          ...(!strictOpenAiCompliance && {
            content_blocks: [
              ...(c.message.reasoning && [{
                type: 'thinking',
                thinking: c.message.reasoning,
              }], 
              {
              type: 'text',
              text: c.message.content
              }
              ),
            ],
          }),
```

please handle similarly for streaming if necessary

unsync: @narengogi i actually made it on my fork but forgot to update the PR 😅 

I initially went for a different idea, because with your snippet i'm adding the content text block *only* if we have a reasoning.

My current solution is this:
```
     choices: response.choices.map((c) => {
        const content_blocks = [];

        if (c.message.reasoning) {
          content_blocks.push({
            type: 'thinking',
            thinking: c.message.reasoning,
          });
        }

        content_blocks.push({
          type: 'text',
          text: c.message.content,
        });

        return {
          index: c.index,
          message: {
            role: c.message.role,
            content: c.message.content,
            ...(!strictOpenAiCompliance && {
              content_blocks
            }),
            ...(c.message.tool_calls && { tool_calls: c.message.tool_calls }),
          },
          finish_reason: c.finish_reason,
        };
      }),
```

That way we always add the text content, and optionally the thinking one, wdyt ?

unsync: it also makes it more scalable if we ever start sending other types of content_block.
by the way, openrouter can return other properties in the responses, depending on the models (like an image_url for some gemini models).

Are you open to allowing other content_block types, like (exemple) `image` or `image_url` ?

narengogi: also, since openrouter documentation does not have complete details on the response object https://openrouter.ai/docs/api-reference/chat-completion ,
are you certain that this is the field in the response? what model are you using?

unsync: @narengogi you can use `anthropic/claude-3.7-sonnet:thinking`.

I agree that their documentation is not great to have a good idea on all responses fields...

You can see examples here : https://openrouter.ai/docs/use-cases/reasoning-tokens#examples

narengogi: got it!

unsync: @narengogi i updated the PR, here is a request sample:

```
--header 'x-portkey-provider: openrouter' \
--header 'x-portkey-config: {"provider": "openrouter","api_key": "sk-or-v1-******"}' \
--header 'x-portkey-strict-open-ai-compliance: false' \
--data '{
    "model": "anthropic/claude-3.7-sonnet:thinking",
    "max_tokens": 2048,
    "thinking": {
        "type": "enabled",
        "budget_tokens": 1024
    },
    "messages": [
        {
            "role": "user",
            "content": "what is 2 + 2 ?"
        }
    ],
    "stream": false
}'
```

```
{
  "id": "gen-1746004839-a8zJrRSyr9NBTQ05sPNQ",
  "object": "chat.completion",
  "created": 1746004839,
  "model": "anthropic/claude-3.7-sonnet:thinking",
  "provider": "openrouter",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "2 + 2 = 4",
        "content_blocks": [
          {
            "type": "thinking",
            "thinking": "This is a simple arithmetic question. 2 + 2 = 4."
          },
          {
            "type": "text",
            "text": "2 + 2 = 4"
          }
        ]
      },
      "finish_reason": "stop"
    }
  ],
  "usage": {
    "prompt_tokens": 45,
    "completion_tokens": 41,
    "total_tokens": 86
  }
}
```

narengogi: yes, your approach looks cleaner
I would only that the content_blocks array be created if strictOpenAiCompliance is false instead of always, you could do
```
     choices: response.choices.map((c) => {
        const content_blocks = [];

      if (!strictOpenAiCompliance) {
        if (c.message.reasoning) {
          content_blocks.push({
            type: 'thinking',
            thinking: c.message.reasoning,
          });
        }

        content_blocks.push({
          type: 'text',
          text: c.message.content,
        });
      }

        return {
          index: c.index,
          message: {
            role: c.message.role,
            content: c.message.content,
            ...(content_blocks.length && {
              content_blocks
            }),
            ...(c.message.tool_calls && { tool_calls: c.message.tool_calls }),
          },
          finish_reason: c.finish_reason,
        };
      }),
```

narengogi (dismissed):
🚀🚀🚀🚀

src/providers/openrouter/chatComplete.ts
unsync: @narengogi that works, updated the PR !

here are samples from a streamed response:

Screenshot 2025-04-30 at 11 23 00
Screenshot 2025-04-30 at 11 23 17

src/providers/openrouter/chatComplete.ts
```diff
@@ -203,6 +237,29 @@ export const OpenrouterChatCompleteStreamChunkTransform: (
...
   const parsedChunk: OpenrouterStreamChunk = JSON.parse(chunk);
+
+  const content_blocks = [];
+  if (!strictOpenAiCompliance) {
+    // add the reasoning first
+    if (parsedChunk.choices?.[0]?.delta?.reasoning) {
+      content_blocks.push({
+        index: parsedChunk.choices?.[0]?.index,
+        delta: {
+          thinking: parsedChunk.choices?.[0]?.delta?.reasoning,
```
VisargD, line 248: Hi, does OpenRouter allow reasoning message in multi-turn conversations? This PR only handles the response mapping but we might also need request mapping to send it in OpenRouter format if it supports the fields in multi-turn conversation.

unsync: @VisargD excellent question !
As far as i know, i want to say they don't handle it: https://openrouter.ai/docs/use-cases/reasoning-tokens

But i never really experimented with multi-turn reasoning, any chance you would have examples of models that handle this ? Or request payloads example ?

Sometimes they (openrouter) have un-documented features depending on the model, so it could be possible

unsync: checked the gateway code, i found what could be the same idea in bedrock, i see their converse api is accepting the reasoningContent in requests, is that what you were referring to ?

Screenshot 2025-05-13 at 16 29 47

If so, i really don't think openrouter supports that as they have (almost) the same signature as openai, which (unless i missed it) doesn't ingest reasoning tokens

VisargD: Yes, It is documented anywhere so there is no way to know the API signature. Will keep an eye on their docs for any updates.

PR #1070, opened 2025-04-29
Fix/support provider specific fields for bedrock in snake case

narengogi:
Code Quality bug fix 

## Author Description

So, portkey node SDK converts input parameters in chat.completions requests into snake_case, when I originally added transforms for `additionalModelRequestFields`, it was in camelCase, it was tested with the REST API and the python sdk, but not the node sdk

# Summary By MatterAI

### 🔄 What Changed
This PR adds support for snake_case versions of provider-specific fields for AWS Bedrock in addition to the existing camelCase versions. The change was necessary because the Portkey Node SDK converts all input parameters in chat.completions requests into snake_case format, which was causing issues with the additionalModelRequestFields and other parameters.

### 🔍 Impact of the Change
This change ensures compatibility with the Portkey Node SDK by supporting both camelCase and snake_case parameter formats. It maintains backward compatibility while fixing the issue with the Node SDK integration.

### 📁 Total Files Changed
2 files changed with 46 additions and 4 deletions:
- src/providers/bedrock/chatComplete.ts: +34, -0
- src/providers/bedrock/utils.ts: +12, -4

### 🧪 Test Added
N/A

### 🔒 Security Vulnerabilities
No security vulnerabilities introduced.

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)

## How Has This Been Tested?
- [ ] Unit Tests
- [ ] Integration Tests
- [x] Manual Testing

## Screenshots (if applicable)
N/A

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

## Related Issues
Issue with Portkey Node SDK converting parameters to snake_case

## Quality Recommendations

1. Consider adding unit tests to verify the snake_case parameter handling works correctly

2. Add documentation comments to explain why both camelCase and snake_case versions are supported

3. Consider creating a utility function to handle the duplicate parameter checking pattern that appears in multiple transform functions

## Sequence Diagram

```mermaid
sequenceDiagram
    participant Client
[...]

PR #1079, opened 2025-05-05
support anthropic_beta as a top level parameter for anthropic on bedr…

narengogi:
support anthropic_beta as a top level parameter for anthropic on bedrock, support both string and array of strings to be compatible with anthropic signature


narengogi:
this PR relates to the changes in https://github.com/Portkey-AI/gateway/pull/1070

VisargD (approved):

PR #1086, opened 2025-05-08
multimodal embeddings for vertex

narengogi:
multimodal embeddings examples

*image*
```json
{
    "model": "multimodalembedding@001",
    "input": [
        {
            "text": "this is the caption of the image",
            "image": {
                "base64": "UklGRkacAABXRUJQVlA4IDqcAACQggKdASqpAn8B....."
            }
        }
    ]
}
```

*video*
```json
{
    "model": "multimodalembedding@001",
    "input": [
        {
            "text": "my screen recording",
            "video": {
                "base64": "AAAAFGZ0eXBxdCAgAA..."
            }
        }
    ]
}
```

### DIFF
--- a/src/providers/google-vertex-ai/transformGenerationConfig.ts
+++ b/src/providers/google-vertex-ai/transformGenerationConfig.ts
@@ -62,6 +62,18 @@ export function transformGenerationConfig(params: Params) {
   return generationConfig;
 }
 
+export function transformEmbeddingsParameters(params: Params) {
+  let embeddingsParameters: Record<string, any> = {};
+  if (params['parameters'] && typeof params['parameters'] === 'object') {
+    embeddingsParameters = { ...params['parameters'] };
+  }
+  if (params['dimensions']) {
+    embeddingsParameters['outputDimensionality'] = params['dimensions'];
+  }
+
+  return embeddingsParameters;
+}
+
 export function transformEmbeddingInputs(params: GoogleEmbedParams) {
   const instances = Array<EmbedInstancesData>();
   if (Array.isArray(params.input)) {
--- a/src/providers/google-vertex-ai/transformGenerationConfig.ts
+++ b/src/providers/google-vertex-ai/transformGenerationConfig.ts
@@ -64,60 +64,38 @@ export function transformGenerationConfig(params: Params) {
 
 export function transformEmbeddingInputs(params: GoogleEmbedParams) {
   const instances = Array<EmbedInstancesData>();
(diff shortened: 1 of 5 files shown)


src/types/embedRequestBody.ts
```diff
@@ -20,8 +37,17 @@ export interface EmbedRequestBody {
...
-  embedding: number[] | number[][]; // The embedding vector(s)
+  embedding?: number[] | number[][]; // The embedding vector(s)
+  image_embedding?: number[];
+  video_embeddings?: {
+    start_offset: number;
+    end_offset?: number;
+    embedding: number[];
+  }[];
   index: number; // The index of the data object
+  type?: string; // The type of data object, e.g., "image", "text", "video"
```
VisargD: - `type` is not required anymore based on the schema introduced in the PR
- `start_offset` and `end_offset` are not part of video_embeddings object so not required on root level.

src/types/embedRequestBody.ts
```diff
@@ -1,9 +1,26 @@
 import { BaseResponse } from '../providers/types';
 import { Options } from './requestBody';
 
+type EmbedInput = {
+  text?: string;
+  image?: {
+    url?: string;
+    base64?: string;
```
VisargD, line 8: base64 seems off from the current nomenclature of the Gateway. Should we instead take base64 or url in the same field like chat completions?

narengogi: I'd prefer to keep it as a separate field, but we can change it if you'd like

src/providers/google-vertex-ai/transformGenerationConfig.ts
```diff
@@ -59,3 +61,54 @@ export function transformGenerationConfig(params: Params) {
...
+    embeddingsParameters = { ...params['parameters'] };
+  }
+  if (params['dimensions']) {
+    embeddingsParameters['outputDimensionality'] = params['dimensions'];
+  }
+
+  return embeddingsParameters;
+}
+
+export function transformEmbeddingInputs(params: GoogleEmbedParams) {
```
VisargD, line 82: dimensions parameter is handled differently for multimodal embeddings. Can you please check if we can integrate that as well?

https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/multimodal-embeddings-api#request_body

narengogi: added

VisargD (approved):

PR #1096, opened 2025-05-14
feat: exclude wrangler configuration from the code

zaherg:
This PR adds `wrangler.toml` and `wrangler.json` to `.gitignore` and provides an example configuration file instead. This prevents merge conflicts that frequently occur when developers modify their local configuration files and then try to pull the latest code.

## Motivation

By excluding these configuration files from version control and providing example templates instead, developers can:
1. Pull new versions without resolving frequent configuration conflicts
2. Maintain their own local configurations without affecting others
3. Enable smoother automation of deployment and testing workflows


docs/installation-deployments.md
```diff
@@ -153,7 +153,13 @@ cd gateway
 npm install
 ```
 
-3. Deploy (using [Wrangler CLI](https://developers.cloudflare.com/workers/wrangler/))
+3. Create a `wrangler` configuration file from the example one we have
+
+```sh
+cp wrangler.toml.example wrangler.toml
```
narengogi, line 159: I would prefer keeping a default wrangler file instead of an example file

narengogi (changes requested):

docs/installation-deployments.md
zaherg: keeping the default file will cause a conflict when editing it by the user and later on by the owner of the code.

This is the whole idea of the PR

PR #1105, opened 2025-05-19
working implementation to enable guardrails for realtime api

narengogi:
Keeping in draft, did this as an experiment to see if we could guardrail realtime audio responses.

This works, to take this to completion
- we'll need to make the checks async
- standardize events to emit for failed guardrail checks that the clients can handle on their end


narengogi:
closing for now

PR #1113, opened 2025-05-27
fix thinking for gemini models

### LINKED ISSUE #1112: Vertex thinking differs from the standardized thinking response from the gateway
### What Happened?

the gateway follows anthropics signature for returning thinking model responses (https://portkey.ai/docs/integrations/llms/anthropic#extended-thinking-reasoning-models-beta)

However this signature is not followed for gemini models, support for which was added prior to the gateway picking up anthropic signature. 
Also google flip flopped between outputting reasoning text and then shutting down reasoning in response and their signature has been slightly modified

_No response_

### Relevant Code Snippet

_No response_

### Your Twitter/LinkedIn

_No response_

narengogi:
closes https://github.com/Portkey-AI/gateway/issues/1112

testing done:
1. tested vertex and google providers with thinking enabled and disabled, in streaming mode and withouth streaming mode
2. verified that caching is working as intended while streaming

example payload:
```json
{
    "model": "gemini-2.5-flash-preview-04-17",
    "max_tokens": 1000,
    "thinking": {
        "budget_tokens": 100,
        "type": "enabled"
    },
    "stream": false,
    "messages": [
        {
            "role": "user",
            "content": [
                {
                    "type": "text",
                    "text": "if tomatina tomatino has fathered tralleliala trallela, and batatina batata is tomatinas sister, how is she related to trallelia?"
                }
            ]
        }
    ]
}
```

### DIFF
--- a/src/providers/google/chatComplete.ts
+++ b/src/providers/google/chatComplete.ts
@@ -481,6 +481,7 @@ interface GoogleGenerateContentResponse {
     promptTokenCount: number;
     candidatesTokenCount: number;
     totalTokenCount: number;
+    thoughtsTokenCount?: number;
   };
 }
 
@@ -583,6 +584,9 @@ export const GoogleChatCompleteResponseTransform: (
         prompt_tokens: response.usageMetadata.promptTokenCount,
         completion_tokens: response.usageMetadata.candidatesTokenCount,
         total_tokens: response.usageMetadata.totalTokenCount,
+        completion_tokens_details: {
+          reasoning_tokens: response.usageMetadata.thoughtsTokenCount ?? 0,
+        },
       },
     };
   }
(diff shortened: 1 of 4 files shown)


csgulati09 (dismissed):

src/providers/google/chatComplete.ts
```diff
@@ -636,35 +647,32 @@ export const GoogleChatCompleteStreamChunkTransform: (
...
+            let content = '';
+            for (const part of generation.content.parts) {
+              if (part.thought) {
+                contentBlocks.push({
+                  index: 0,
+                  delta: { thinking: part.text },
+                });
+                streamState.containsChainOfThoughtMessage = true;
+              } else {
+                content = part.text ?? '';
```
b4s36t4, line 669: Here we're overriding the variable value, is this expected? shouldn't we append the `part.text`.

narengogi: yes this is expected, we intialize the value to an empty string and if strict openai compliance is false, we set it to the inference message

VisargD (approved):

PR #1171, opened 2025-06-25
feat: add support for vertex-ai global region endpoints

### LINKED ISSUE #1172: Vertex global region endpoints
Vertex global region does not require region in the base url. It should be handled in the Gateway as well.

Vertex documentation: https://cloud.google.com/vertex-ai/generative-ai/docs/learn/locations#use_the_global_endpoint

VisargD:
## Type of Change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Testing

## Checklist

## Related Issues
Closes #1172

### DIFF
--- a/src/providers/google-vertex-ai/api.ts
+++ b/src/providers/google-vertex-ai/api.ts
@@ -57,6 +57,9 @@ export const GoogleApiConfig: ProviderAPIConfig = {
       return `https://storage.googleapis.com`;
     }
 
+    if (vertexRegion === 'global') {
+      return `https://aiplatform.googleapis.com`;
+    }
     return `https://${vertexRegion}-aiplatform.googleapis.com`;
   },
   headers: async ({ c, providerOptions }) => {


b4s36t4 (approved):

src/providers/google-vertex-ai/api.ts
```diff
@@ -57,6 +57,9 @@ export const GoogleApiConfig: ProviderAPIConfig = {
       return `https://storage.googleapis.com`;
     }
 
+    if (vertexRegion === 'global') {
+      return `https://aiplatform.googleapis.com`;
```
narengogi, line 61: just adding as a note, I've just tested this endpoint with normal regions like `us-east5` as well and it works, so I think we can make this as the default instead of ln:63

VisargD: Based on their docs, it only works for some models. So partner models like anthropic and meta models might not work if we change the url.
https://cloud.google.com/vertex-ai/generative-ai/docs/learn/locations#supported_models

PR #1175, opened 2025-06-25
Integration: Bytez Chat Model Provider

inf3rnus:
## Description
Hi all! :wave: 

We're Bytez, the largest model provider on the internet! We may also be one of the cheapest if not the cheapest.

We'd love to integrate with PortKey. Please see the changed files and let me know if anything needs to change.

I'd like to point out that we do a check against our api to see if a model is a "chat" model. This is stored in a simple cache that is just an object. If that's going to be a problem due to having an unbounded ceiling in terms of memory utilization pls lmk, and I will convert it to a LRU with 100 entries.

Our API's input signature is a bit more bespoke than other model providers, please lmk if the custom requestHandler I have is sufficient, or if there's an easier way to do what I've done.

Bonus feedback: Ya'll need an integration guide! Would be immensely useful :smile:

## Motivation
We'd love to be integrated in PortKey!

## Type of Change

- [x ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## How Has This Been Tested?

- [ ] Unit Tests
- [ ] Integration Tests
- [ x] Manual Testing

## Checklist

- [ x] My code follows the style guidelines of this project
- [ x] I have performed a self-review of my own code
- [ x] I have commented my code, particularly in hard-to-understand areas
- [ x] My changes generate no new warnings


src/providers/bytez/index.ts
```diff
@@ -0,0 +1,234 @@
+import crypto from 'node:crypto';
+import { ParameterConfig, ProviderConfigs } from '../types';
+import BytezInferenceAPI from './api';
+import { BytezInferenceChatCompleteConfig } from './chatComplete';
+
+const BASE_URL = 'https://api.bytez.com/models/v2';
+
+const IS_CHAT_MODEL_CACHE: Record<string, boolean> = {};
```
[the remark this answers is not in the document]
inf3rnus: Added a LRU instead

src/providers/bytez/index.ts
```diff
@@ -0,0 +1,234 @@
...
+    output: [model],
+  }: BytezResponse = await response.json();
+
+  if (error) {
+    throw new Error(error);
+  }
+
+  const isChatModel = model.task === 'chat';
+
+  IS_CHAT_MODEL_CACHE[modelId] = isChatModel;
```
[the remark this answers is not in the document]
inf3rnus: Added a LRU

src/providers/bytez/index.ts
```diff
@@ -0,0 +1,234 @@
...
+  // add other model properties as needed
+}
+
+async function validateModelIsChat(
+  modelId: string,
+  headers: Record<string, any>
+) {
+  // return from cache if already validated
+  if (IS_CHAT_MODEL_CACHE[modelId]) {
+    return IS_CHAT_MODEL_CACHE[modelId];
```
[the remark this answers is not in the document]
inf3rnus: Added LRU instead.

src/providers/bytez/index.ts
```diff
@@ -0,0 +1,234 @@
...
+  if (error) {
+    throw new Error(error);
+  }
+
+  const isChatModel = model.task === 'chat';
+
+  IS_CHAT_MODEL_CACHE[modelId] = isChatModel;
+
+  return isChatModel;
+}
```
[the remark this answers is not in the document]
inf3rnus: Added a wrapping try catch block to always send a failure response to the client

inf3rnus:
```
Consider replacing the custom LRUCache implementation with a well-tested, battle-hardened library for caching to improve robustness and reduce potential maintenance overhead.
```
We do not want to bloat the code base with packages that aren't needed. It should be sufficient.

```
Enhance the constructFailureResponse function to map specific Bytez API error codes (e.g., 4xx client errors, 401 authentication errors) to appropriate HTTP status codes instead of defaulting to a generic 500. This provides more informative error responses to the client.
```
It already does this where necessary, otherwise whatever is passed from the server is passed to the client

```
Review the validateModelIsChat function's error handling. While it currently throws an error that is caught, consider if it should directly return a Response object using constructFailureResponse for consistency with other API error paths.
```
Overkill, either fetch is going to fail, or there is an upstream error, either way it will get reported to the client.

```
Add comprehensive unit tests for the new Bytez integration logic, specifically for bodyAdapter, LRUCache, and the chatComplete request handler, to ensure correct functionality and prevent regressions.
```
Perhaps in the future...

```
Verify the splitPattern = ' ' for Bytez in getStreamModeSplitPattern against Bytez's actual streaming API documentation to confirm it correctly handles their specific stream format, as this is an unusual pattern for streaming.
```

We currently stream character by character, we do not return JSON chunks. In the future we may update this.

```
Explore making the bodyAdapter logic more declarative or configurable, especially if future provider integrations might require similar but slightly varied request body transformations. This could improve maintainability and readability for complex adaptations.
```
YAGNI and overkill, bytez manages its own integration code.

src/providers/bytez/types.ts
```diff
@@ -0,0 +1,11 @@
+interface Model {
+  task: string;
+}
+
+interface BytezResponse {
+  error: string;
+  output: Model[];
+  // add other model properties as needed
```
narengogi: please remove comment if not required

src/handlers/handlerUtils.ts
```diff
@@ -510,7 +510,12 @@ export async function tryPost(
           body: transformedRequestBody,
           headers: fetchOptions.headers,
         },
-        requestParams: transformedRequestBody,
+        requestParams: {
```
narengogi: please remove these changes if you did not intend them, why are you merging request params to transformedRequestBody?

src/providers/bytez/index.ts
```diff
@@ -0,0 +1,130 @@
+import crypto from 'node:crypto';
```
narengogi: please remove this import, we use the inbuilt crypto module to support deployment across different runtimes

src/providers/bytez/utils.ts
```diff
@@ -0,0 +1,129 @@
+import { ParameterConfig } from '../types';
```
narengogi: caching is provided by portkey through `putInCacheWithValue`  and `getFromCacheByKey` which are set in context, please remove this

src/providers/bytez/index.ts
```diff
@@ -0,0 +1,130 @@
...
+import { BytezResponse } from './types';
+
+const BASE_URL = 'https://api.bytez.com/models/v2';
+
+const IS_CHAT_MODEL_CACHE = new LRUCache({ size: 100 });
+
+const BytezInferenceAPIConfig: ProviderConfigs = {
+  api: BytezInferenceAPI,
+  chatComplete: BytezInferenceChatCompleteConfig,
+  requestHandlers: {
```
narengogi: you do not need both chatCompleteConfig and a requestHandler, I'd suggest you remove the requestHandler because this logic is not for the gateway to handle
plus you're missing the response transforms

narengogi (changes requested):

src/providers/bytez/index.ts
inf3rnus: done!

src/providers/bytez/types.ts
inf3rnus: Done!

src/providers/bytez/utils.ts
inf3rnus: Hey @narengogi 

Took a look at the implementation in `src/middlewares/cache/index.ts`, looks like it needs to be enabled in conf.json for it to be made accessible within the hono Context, for our integration we always want the cache on because users will get a degraded experience otherwise. (Poor throughput)... We could just forgo the cache entirely, but it will be a bad DX, i.e. user goes to use an unsupported model and then bam, error message, but not one that mentally maps easily to "this model does not support chat."

One other issue is that the impl in `src/middlewares/cache/index.ts` doesn't support an eviction strategy, so its size is unbounded, which is going to be bad for anyone running PortKey in prod.

What are your thoughts on allowing us to keep our LRU implementation?

My philosophy with BYO adapter projects is that as long as no additional packages are required, functionality is the responsibility of the upstream api owner.

src/handlers/handlerUtils.ts
inf3rnus: yeah these changes were intentional, transformedRequestBody is going to be an empty {} if a `requestHandler` is used instead of a `responseTransformer`

and then the consequence is you do not get the contents of the request from `processLog` (Code block below)

The ideal is we do not use requestHandler, but we desire a greater deal of flexibility than what's currently provided, see the comments of the PR for thoughts on how we can enhance the DX as a model provider!

```js
async function processLog(c: Context, start: number) {
  const ms = Date.now() - start;
  if (!c.req.url.includes('/v1/')) return;

  const requestOptionsArray = c.get('requestOptions');
  if (!requestOptionsArray?.length) {
    return;
  }

  try {
    const response = requestOptionsArray[0].requestParams.stream
      ? { message: 'The response was a stream.' }
      : await c.res.clone().json();

    const responseString = JSON.stringify(response);
    if (responseString.length > MAX_RESPONSE_LENGTH) {
      requestOptionsArray[0].response =
        responseString.substring(0, MAX_RESPONSE_LENGTH) + '...';
    } else {
      requestOptionsArray[0].response = response;
    }
  } catch (error) {
    console.error('Error processing log:', error);
  }

  await broadcastLog(
    JSON.stringify({
      time: new Date().toLocaleString(),
      method: c.req.method,
      endpoint: c.req.url.split(':8787')[1],
      status: c.res.status,
      duration: ms,
      requestOptions: requestOptionsArray,
    })
  );
}async function processLog(c: Context, start: number) {
  const ms = Date.now() - start;
  if (!c.req.url.includes('/v1/')) return;

  const requestOptionsArray = c.get('requestOptions');
  if (!requestOptionsArray?.length) {
    return;
  }

  try {
    const response = requestOptionsArray[0].requestParams.stream
      ? { message: 'The response was a stream.' }
      : await c.res.clone().json();

    const responseString = JSON.stringify(response);
    if (responseString.length > MAX_RESPONSE_LENGTH) {
      requestOptionsArray[0].response =
        responseString.substring(0, MAX_RESPONSE_LENGTH) + '...';
    } else {
      requestOptionsArray[0].response = response;
    }
  } catch (error) {
    console.error('Error processing log:', error);
  }

  await broadcastLog(
    JSON.stringify({
      time: new Date().toLocaleString(),
      method: c.req.method,
      endpoint: c.req.url.split(':8787')[1],
      status: c.res.status,
      duration: ms,
      requestOptions: requestOptionsArray,
    })
  );
}
```

src/providers/bytez/index.ts
inf3rnus: Hey @narengogi with our impl we do need that config, we end up using it to adapt the params ourselves in requestHandler via the func we wrote `bodyAdapter` (see the code block below). The intention is that we will end up using it in a way that better fits the expectations of the code base.

The reason we aren't using `responseTransformer` is because the existing config impl is not flexible enough yet for us to use it, our API has a different input signature than the OpenAI compliant standard supports, e.g. our request body looks like this:

```
{
"messages" [],
"stream": true | false,
"params": {}
}
```

We need to lift params that would normally be at level 1 of the req.body's object and move them into a dedicated prop called `params`.

To provide a higher degree of flexibility I think the ProviderConfig interface should be extended to support requestTransformers so that you can use it for `fn`'s `chatComplete` et. al, currently it can only be used with fileUpload etc.

The idea would be that we modify `transformToProviderRequest` so that if you specify `chatCompletion` under `requestTransformers` you can specify how you want your request body to be transformed for `chatCompletion`.

Right now it's hard locked to just remapping the names of the params with `transformUsingProviderConfig`.

```js
function bodyAdapter(requestBody: Record<string, any>) {
  for (const [param, paramConfig] of Object.entries(
    BytezInferenceChatCompleteConfig
  )) {
    const hasParam = Boolean(requestBody[param]);

    // first assign defaults
    if (!hasParam) {
      const { default: defaultValue, required } =
        paramConfig as ParameterConfig;

      // if it's required, throw
      if (required) {
        throw new Error(`Param ${param} is required`);
      }

      // assign the default value
      if (defaultValue !== undefined && requestBody[param] === undefined) {
        requestBody[param] = defaultValue;
      }
    }
  }

  // now we remap everything that has an alias, i.e. "prop" on propConfig
  for (const key of Object.keys(requestBody)) {
    const paramObj = BytezInferenceChatCompleteConfig[key] as
      | ParameterConfig
      | undefined;

    if (paramObj) {
      const { param: alias } = paramObj;

      if (key !== alias) {
        requestBody[alias] = requestBody[key];
        delete requestBody[key];
      }transformUsingProviderConfig
    }
  }

  // now we adapt to the bytez input signature
  // props to skip
  const skipProps: Record<string, boolean> = {
    model: true,
  };

  // props that cannot be removed from the body
  const reservedProps: Record<string, boolean> = {
    stream: true,
    messages: true,
  };

  const adaptedBody: Record<string, any> = { params: {} };

  for (const [key, value] of Object.entries(requestBody)) {
    // things like "model"
    if (skipProps[key]) {
      continue;
    }

    // things like "messages", "stream"
    if (reservedProps[key]) {
      adaptedBody[key] = value;
      continue;
    }
    // anything else, e.g. max_new_tokens
    adaptedBody.params[key] = value;
  }

  return adaptedBody;
}
```

src/providers/bytez/utils.ts
narengogi: @inf3rnus this is poor design pattern, caching is not provider specific. It is for the gateway to handle centrally, which is how it is currently designed. please remove this from here

src/providers/bytez/index.ts
```diff
@@ -0,0 +1,129 @@
...
+          response
+        );
+      } catch (error: any) {
+        return constructFailureResponse(error.message);
+      }
+    },
+  },
+};
+
+async function validateModelIsChat(
```
VisargD: Hi! Adding too much provider-specific logic would make Gateway harder to maintain. Ideally, the Bytez chat API should return an error if an invalid model is passed.

For example, Gateway does not maintain a list of valid models for OpenAI or make decisions on behalf of OpenAI. It simply maps the model name and forwards the request, letting the provider handle validation. 

Performing a separate model validation request before making the actual API call is redundant, since the provider’s API should return an error for invalid models anyway.

src/providers/bytez/index.ts
VisargD: > We need to lift params that would normally be at level 1 of the req.body's object and move them into a dedicated prop called params

Hey, the default request transformer is already flexible to handle this use-case. You can add a `transform` parameter in the ChatComplete config which uses a function to do the required transformations. Example:

```
encoding_format: {
    param: 'embedding_types',
    required: false,
    transform: (params: any): string[] => {
     ...logic
    },
  },
```

We already have providers (Example: vertex-ai) which already moves props from root level to some nested object based on the API signature.

inf3rnus:
@narengogi @VisargD Requested changes have been made, and I dropped the LRU. 

Pls lmk if you need any further changes.

src/providers/bytez/index.ts
```diff
@@ -0,0 +1,56 @@
...
+        model: requestBody.model,
+        choices: [
+          {
+            index: 0,
+            message: output,
+            logprobs: null,
+            finish_reason: 'stop',
+          },
+        ],
+        usage: {
```
narengogi: please make the usage object openai compliant
https://portkey.ai/docs/api-reference/inference-api/chat#response-usage

narengogi: also please move this function into `chatComplete.ts`

narengogi (changes requested):
looks good to go! adding two minor comments

src/providers/bytez/index.ts
inf3rnus: @narengogi Alright, changes have been made! 

We currently do not provide token usage metrics as part of our API (support for it will be added soon, although our billing model is a bit unique in that it's based around request concurrency).

Following other examples where the counts are not returned, I did this, this work for you guys?

If it's a hard requirement lmk, and I'll update our backend ASAP. 

```js
    usage: {
      completion_tokens: -1,
      prompt_tokens: -1,
      total_tokens: -1,
    },
```

inf3rnus:
@narengogi @VisargD Pinging for visibility, we should be ready to rock :smile:

narengogi (approved):

inf3rnus:
@narengogi  @VisargD Hey again folks!

We've created the docs that correspond to this pr on your docs repos, very much appreciate you taking the time to look this all over :smile: 

https://github.com/Portkey-AI/docs-core/pull/455

PR #1228, opened 2025-07-17
improvement: updated ai21 integration

### LINKED ISSUE #1215: Update ai21 integration
### What Would You Like to See with the Gateway?

the ai21 integration is pretty dated, the existing integration makes calls to `/v1/{model}/chat` whereas the new API is `/v1/chat/completions`, the endpoint is mostly openai compliant
reference: https://docs.ai21.com/reference/jamba-1-6-api-ref

### Context for your Request

_No response_

### Your Twitter/LinkedIn

_No response_

Utkarsh-0304:
## Description

Updated Chat Completions API endpoint, and also modified API configs and response interface

## Motivation

This updates the existing outdated ai21 integration and fixes #1215 

## Type of Change

- [ ] Documentation update

## Checklist

## Related Issues
Fixes #1215

### DIFF
--- a/src/providers/ai21/chatComplete.ts
+++ b/src/providers/ai21/chatComplete.ts
@@ -1,5 +1,5 @@
 import { AI21 } from '../../globals';
-import { Params, SYSTEM_MESSAGE_ROLES } from '../../types/requestBody';
+import { Params } from '../../types/requestBody';
 import {
   ChatCompletionResponse,
   ErrorResponse,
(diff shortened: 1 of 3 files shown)


src/providers/ai21/chatComplete.ts
```diff
@@ -47,77 +51,69 @@ export const AI21ChatCompleteConfig: ProviderConfig = {
...
   n: {
     param: 'numResults',
     default: 1,
+    min: 1,
+    max: 16,
   },
-  max_tokens: {
-    param: 'maxTokens',
-    default: 16,
+  documents: {
```
narengogi: you would need to handle mapping from user message content parts containing files/images or audio (whichever is supported by ai21) to the documents field

Image

so for the fixed mapping you need to move this inside the `messages` object like this

```json
  messages: [{param: 'messages'...}, {param: 'documents' ...}]
```

narengogi: in the new ai21 spec, there is no separate field for 'system', so you can remove this code block
```json
    {
      param: 'system',
      required: false,
      transform: (params: Params) => {
        if (
          params.messages?.[0]?.role &&
          SYSTEM_MESSAGE_ROLES.includes(params.messages?.[0]?.role)
        ) {
          return params.messages?.[0].content;
        }
      },
    },
```

narengogi: also remove this chunk from the messages object
```js
if (
          params.messages?.[0]?.role &&
          SYSTEM_MESSAGE_ROLES.includes(params.messages?.[0]?.role)
        ) {
          inputMessages = params.messages.slice(1);
```

narengogi: you will also need to map the text content part into the user and system messages from openai to ai21 format
in openai, a user can send user message like this
```
{
"role": "user",
"content" : [{"type": "text", "text": "abc"}, ..multiple parts]
}
```
but ai21 only accepts a string, so you will need to handle that

narengogi (changes requested):

src/providers/ai21/chatComplete.ts
Utkarsh-0304: Thanks for your review! I'm working on the requested changes and will update the pull request soon.

Also, could you please send me the link for the above-attached image for reference? As in the official [API reference for AI21 Chat completions](https://docs.ai21.com/reference/jamba-1-6-api-ref), parameters are defined similarly to the ones in my commit.

PR #1239, opened 2025-07-23
onStatusCodes is respected during fallbacks

roh26it:
## Description
onStatusCodes is now respected during fallbacks

### DIFF
--- a/src/handlers/handlerUtils.ts
+++ b/src/handlers/handlerUtils.ts
@@ -671,7 +671,7 @@ export async function tryTargetsRecursively(
           break;
         }
         if (
-          response?.ok &&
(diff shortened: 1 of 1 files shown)


src/handlers/handlerUtils.ts
```diff
@@ -671,7 +671,7 @@ export async function tryTargetsRecursively(
           break;
         }
         if (
-          response?.ok &&
+          currentTarget.strategy?.onStatusCodes?.length &&
```
VisargD: If onStatusCodes is not sent by the user, we should still fallback for all non-2xx status codes same as the existing logic.

VisargD (approved):

PR #1121, opened 2025-06-03
refactored tryPost

roh26it:
## Description
Cleaning up handlerUtils in preparation to add more features here.

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [x ] Refactoring (no functional changes)

## How Has This Been Tested?
- [ ] Unit Tests
- [ ] Integration Tests
- [x ] Manual Testing

## Checklist

- [x ] My code follows the style guidelines of this project
- [ x] I have performed a self-review of my own code
- [ x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #1241, opened 2025-07-24
fix: handle provider error logs addition

VisargD:
## Description
logObject.log() method is not getting called for error responses because responseService.create() throws an error with response details for all non-2xx LLM error responses. This causes issues with keeping track of logObjects of failed requests. 

It is redundant to throw the error from responseService, because it is simply getting caught and then the same value is used in the final response object of tryTargetsRecursively function. 

With this change, tryTargetsRecursively catch block should only be used when there is an actual gateway exception. For all provider non-2xx responses, they will be treated as a valid response. 

## Motivation
Fix on top of #1121

## Type of Change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Testing

## Checklist

### DIFF
--- a/src/handlers/handlerUtils.ts
+++ b/src/handlers/handlerUtils.ts
@@ -790,53 +790,35 @@ export async function tryTargetsRecursively(
         // tryPost always returns a Response.
         // TypeError will check for all unhandled exceptions.
         // GatewayError will check for all handled exceptions which cannot allow the request to proceed.
-        if (
-          error instanceof TypeError ||
-          error instanceof GatewayError ||
-          !error.response ||
-          (error.response && !(error.response instanceof Response))
-        ) {
-          console.error(
-            'tryTargetsRecursively error: ',
-            error.message,
-            error.cause,
-            error.stack
-          );
-          const errorMessage =
-            error instanceof GatewayError
-              ? error.message
-              : 'Something went wrong';
-          response = new Response(
-            JSON.stringify({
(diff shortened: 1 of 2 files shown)


src/handlers/services/responseService.ts
```diff
@@ -62,13 +62,6 @@ export class ResponseService {
 
     this.updateHeaders(finalMappedResponse, cache.cacheStatus, retryAttempt);
 
-    if (!finalMappedResponse.ok) {
```
VisargD, line 65: Instead of throwing error, tryPost will return it just as a simple response as throwing error is redundant.

In tryTargetsRecursively, we are already checking for response.ok

src/handlers/handlerUtils.ts
```diff
@@ -790,53 +790,35 @@ export async function tryTargetsRecursively(
...
-            {
-              status: 500,
-              headers: {
-                'content-type': 'application/json',
-                // Add this header so that the fallback loop can be interrupted if its an exception.
-                'x-portkey-gateway-exception': 'true',
-              },
-            }
-          );
-        } else {
```
VisargD, line 828: Since this PR removes the throw new Error() logic in tryPost for LLM failures, we no longer need to set it here. If the any error is caught, they will always be an Gateway exception.

narengogi (approved):

Issue #1217, opened 2025-07-07
Audio transcription endpoint does not respect Gateway's override_params


AgileEduLabs:
### What Happened?

With `portkey.chat.completions.create`, we can provide `override_params` to the gateway to specify which model to use. The `model` parameter in `portkey.chat.completions.create` itself can be omitted since we specify it in the gateway.

However, with `portkey.audio.transcriptions.create`, there are **two** issues.

**First**, it forces us to specify the model. Otherwise it will cause an error:
```
portkey.audio.transcriptions.create({
          model: '', // <--- Model parameter must be specified. If omitted, it will cause an error
          file: fs.createReadStream(fileName),
          response_format: 'text',
        }, requestOptionsForTranscription);
```

**Second**, even more problematic, `portkey.audio.transcriptions.create` does not respect the `override_params` of the gateway config object.
In the snipped I have included below, we would expect it to use `distil-whisper-large-v3-en` as specified in the config object. However, this value is ignored and instead portkey uses the non-existent `should-use-model-name-from-gateway` value.

### Relevant Code Snippet

        const portkeyGatewayConfigForTranscription = {
          "strategy": {
            "mode": "loadbalance"
          },
          "targets": [
            {
              "strategy": {
                "mode": "fallback"
              },
              "targets": [
                {
                  "provider": "groq",
                  "api_key": "gsk_",
                  "override_params": {
                    "model": "distil-whisper-large-v3-en"
                  }
                },
              ]
            },
            {
              "strategy": {
                "mode": "fallback"
              },
              "targets": [
                {
                  "provider": "groq",
                  "api_key": "gsk_",
                  "override_params": {
                    "model": "distil-whisper-large-v3-en"
                  }
                },
              ]
            }
          ]
        }

        const requestOptionsForTranscription = {
          metadata: {
          },
          config: JSON.stringify(portkeyGatewayConfigForTranscription),
        }
        let transcript = await portkey.audio.transcriptions.create({
          model: 'should-use-model-name-from-gateway',
          file: fs.createReadStream(fileName),
          response_format: 'text',
        }, requestOptionsForTranscription);

narengogi:
this is expected @AgileEduLabs , override_params is for json parameters,
it is unfeasible for a gateay to load a `multipart/form-data` payload into memory and transform the parameters

AgileEduLabs:
Hi @narengogi , it seems that data is encoded to `multipart/form-data` on the gateway itself like so:

```
[...]

PR #1242, opened 2025-07-25
fix: apply override_params to FormData request bodies

AgileEduLabs:
## Description
Transform the javascript FormData object based on the supplied override_params. Modified transformToProviderRequest.ts on line 235:
```
  if (requestBody instanceof FormData) {
    if (typeof providerOptions.overrideParams === 'undefined') {
      return requestBody;
    }

    for (const [k, v] of Object.entries(providerOptions.overrideParams)) {
      requestBody.delete(k); // If the key already exists, delete it first so we always overwrite.

      // Value can be string, Blob, File, or an array of those.
      const values = Array.isArray(v) ? v : [v];
      for (const val of values) {
        requestBody.append(k, val);
      }
    }

    return requestBody;

  }
```

## Motivation
Currently `portkey.audio.transcriptions.create` does not respect the `override_params` of the gateway config object. For example, if we change the model name in `override_params`, we would expect it to use the model name as specified in the config object. However, this values in the config object are ignored.

If override_params cannot modify the model name on the gateway, then the gateway features are essentially useless. You won't be able to do any sort of meaningful fallback nor loadbalancing, because each provider has their own model name. 

The source of the problem lies in transformToProviderRequest.ts on line 235. If body is FormData, then override parameters are not applied and instead the function returns early with the requestBody unmodified:
```
if (requestBody instanceof FormData || requestBody instanceof ArrayBuffer)
    return requestBody;
```

At first glance this makes sense because you cannot apply a simple spread operator to FormData nor ArrayBuffer, so we just ignore the override_params and return early.

However, the more elegant solution is to transform the javascript FormData object based on the override_params. Since requestBody is **already** a JavaScript FormData object, the computational cost of overriding the keys is **negligible**. This is in keeping with the spirit of the gateway's lightweight nature:

```
  for (const [k, v] of Object.entries(providerOptions.overrideParams)) {
      requestBody.delete(k); // If the key already exists, delete it first so we always overwrite.

      // Value can be string, Blob, File, or an array of those.
      const values = Array.isArray(v) ? v : [v];
      for (const val of values) {
        requestBody.append(k, val);
      }
    }
```

This pull request is to fix the issue as outlined in [https://github.com/Portkey-AI/gateway/issues/1217](https://github.com/Portkey-AI/gateway/issues/1217)

## Type of Change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Manual Testing

## Checklist

- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/src/services/transformToProviderRequest.ts
+++ b/src/services/transformToProviderRequest.ts
@@ -232,7 +232,25 @@ export const transformToProviderRequest = (
       fn
     );
   }
-  if (requestBody instanceof FormData || requestBody instanceof ArrayBuffer)
+
+  if (requestBody instanceof FormData) {
+    if (typeof providerOptions.overrideParams === 'undefined') {
+      return requestBody;
+    }
+
+    for (const [k, v] of Object.entries(providerOptions.overrideParams)) {
+      requestBody.delete(k); // If the key already exists, delete it first so we always overwrite.
+
+      // Value can be string, Blob, File, or an array of those.
+      const values = Array.isArray(v) ? v : [v];
+      for (const val of values) {
+        requestBody.append(k, val);
+      }
+    }
+
+    return requestBody;
+
+  } else if (requestBody instanceof ArrayBuffer)
     return requestBody;
 
   if (fn === 'proxy') {


narengogi:
@AgileEduLabs I think we need to take a more nuanced approach with regards to the tts and stt routes unification, though we currently support loading and transforming multipart form data for these routes, It is not scalable, I want to verify if we can transform these on the fly with stream transforms, similar to how we are doing for file uploads

PR #1279, opened 2025-08-12
[vertex-ai] fix: allow object as tool content

arturfromtabnine:
## Description

Allow tool content to be a typeof object for google models on vertex (which is compatible with open ai api https://github.com/openai/openai-node/blob/40b79cb8a372dce925b14a26e62110f9de6786ba/src/resources/chat/completions/completions.ts#L1407). Otherwise if content is a typeof object, `role` would not be handled properly and request will fail with following error:
```
{"error":{"message":"vertex-ai error: Please use a valid role: user, model.","type":"INVALID_ARGUMENT","param":null,"code":"400"},"provider":"vertex-ai"}
```

https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling 
## Type of Change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Manual Testing

## Checklist

### DIFF
--- a/src/providers/google-vertex-ai/chatComplete.ts
+++ b/src/providers/google-vertex-ai/chatComplete.ts
@@ -52,6 +52,18 @@ import {
   transformVertexLogprobs,
 } from './utils';
 
+const isContentTypeArray = (content: any): content is ContentType[] => {
+  return (
+    Array.isArray(content) &&
+    content.every(
+      (item) =>
+        typeof item === 'object' &&
+        item !== null &&
+        typeof item.type === 'string'
+    )
+  );
+};
+
 export const buildGoogleSearchRetrievalTool = (tool: Tool) => {
   const googleSearchRetrievalTool: GoogleSearchRetrievalTool = {
     googleSearchRetrieval: {},
@@ -102,7 +114,7 @@ export const VertexGoogleChatCompleteConfig: ProviderConfig = {
           } else if (
             message.role === 'tool' &&
             (typeof message.content === 'string' ||
-              typeof message.content === 'object')
+              isContentTypeArray(message.content))
           ) {
             parts.push({
               functionResponse: {
--- a/src/providers/google-vertex-ai/chatComplete.ts
+++ b/src/providers/google-vertex-ai/chatComplete.ts
@@ -101,7 +101,8 @@ export const VertexGoogleChatCompleteConfig: ProviderConfig = {
             });
           } else if (
             message.role === 'tool' &&
-            typeof message.content === 'string'
+            (typeof message.content === 'string' ||
+              typeof message.content === 'object')
           ) {
             parts.push({
               functionResponse: {
--- a/src/providers/google-vertex-ai/chatComplete.ts
+++ b/src/providers/google-vertex-ai/chatComplete.ts
@@ -52,18 +52,6 @@ import {
   transformVertexLogprobs,
 } from './utils';
 
-const isContentTypeArray = (content: any): content is ContentType[] => {
-  return (
-    Array.isArray(content) &&
-    content.every(
-      (item) =>
-        typeof item === 'object' &&
-        item !== null &&
-        typeof item.type === 'string'
-    )
-  );
-};
-
 export const buildGoogleSearchRetrievalTool = (tool: Tool) => {
   const googleSearchRetrievalTool: GoogleSearchRetrievalTool = {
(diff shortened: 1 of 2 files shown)


src/providers/google-vertex-ai/chatComplete.ts
```diff
@@ -101,7 +101,8 @@ export const VertexGoogleChatCompleteConfig: ProviderConfig = {
             });
           } else if (
             message.role === 'tool' &&
-            typeof message.content === 'string'
+            (typeof message.content === 'string' ||
+              typeof message.content === 'object')
```
narengogi: OpenAI spec only supports sending text content or array of text content parts as part of tool message content https://platform.openai.com/docs/api-reference/chat/create#chat_create-messages

I believe what you are trying to do is to send an object like this 
```json
{
  "role": "user",
  "parts": [
    {
      "functionResponse": {
        "name": "get_current_weather",
        "response": {
          "temperature": 20,
          "unit": "C"
        }
      }
    }
  ]
}
```
instead of 
```json
{
  "role": "user",
  "parts": [
    {
      "functionResponse": {
        "name": "get_current_weather",
        "response": { "content": "{'temperature': 20, 'unit': 'C'}"}
      }
    }
  ]
}
```

the current implementation is definitely incomplete, what we should be doing is
```js
 else if (
            message.role === 'tool' &&
            typeof message.content === 'string'
          ) {
            let toolResponse;
            try {
              toolResponse = JSON.parse(message.content);
            } catch (e) {
              toolResponse = {
                content: message.content,
              };
            }
            parts.push({
              functionResponse: {
                name: message.name ?? 'gateway-tool-filler-name',
                response,
              },
            });
          } 
```

narengogi: we need to handle array of text contents as well, which we can change in a separate PR

narengogi (changes requested):
Hey @arturfromtabnine 
Openai chat completion [content is always a string](https://platform.openai.com/docs/api-reference/chat/object#chat/object-choices)

src/providers/google-vertex-ai/chatComplete.ts
arturfromtabnine: You're right, we need to do a strict check for array, but your example is not what we are trying to send. Our client is open ai compatible, so its trying to send an `array` in `message.content` 
```
{
role: 'tool',
tool_call_id: 'call_ty2u3gJiEIsf9TLOwUB67opO',
content: [
  {
   type: 'text',
   text: '{"filePath"}'
  }
 ]
}
```
with suggested change there no issue, vertex api for gemini models allows to have response.content as an array or string

arturfromtabnine: Not sure if we need to parse content like was suggested
```
            typeof message.content === 'string'
          ) {
            let toolResponse;
            try {
              toolResponse = JSON.parse(message.content);
            } catch (e) {
              toolResponse = {
                content: message.content,
              };
            }
```

arturfromtabnine:
> Hey @arturfromtabnine Openai chat completion [content is always a string](https://platform.openai.com/docs/api-reference/chat/object#chat/object-choices)

Hey @narengogi  it can be an array as well 
Screenshot 2025-08-13 at 16 50 54

narengogi:
> > Hey @arturfromtabnine Openai chat completion [content is always a string](https://platform.openai.com/docs/api-reference/chat/object#chat/object-choices)
> 
> Hey @narengogi it can be an array as well

But it's an array of text content parts, you need an object for vertex right?

arturfromtabnine:
> > > Hey @arturfromtabnine Openai chat completion [content is always a string](https://platform.openai.com/docs/api-reference/chat/object#chat/object-choices)
> > 
> > 
> > Hey @narengogi it can be an array as well
> 
> But it's an array of text content parts, you need an object for vertex right?

Sorry, was off
we do send content for tools in form of array , not an object
```javascript
        content: [
          {
            type: "text",
            text: "content",
          },
        ],
  ``` 
  will update suggested condition to be more precise on array instead of being too generic

narengogi (approved):
I've removed the check entirely and updated the key from `content` to `output` as mentioned in the vertex spex
Image

https://cloud.google.com/vertex-ai/docs/reference/rest/v1/Content#FunctionResponse

VisargD (approved):

PR #1296, opened 2025-08-21
integration: add tripo3d provider

dd-eg-user:
## Description

  Adds comprehensive Tripo3D provider integration to the Portkey AI Gateway, enabling users to route 3D model generation requests through Portkey to Tripo3D's API. This implementation provides full support for all Tripo3D endpoints
  including task creation, status polling, file uploads, and account management.
## Motivation

  Tripo3D is a leading 3D generation platform that converts text prompts and images into high-quality 3D models. Adding Tripo3D as a provider expands Portkey's capabilities into the 3D generation space, allowing users to leverage Portkey's
   reliability features (retries, fallbacks, load balancing) with Tripo3D's powerful 3D AI models.
## Type of Change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Manual Testing

## Checklist

- [ x I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes


roh26it:
@dd-eg-user as discussed, would you be able to add `credits_used` to the response object to enable pricing configs?

dd-eg-user:
> @dd-eg-user as discussed, would you be able to add `credits_used` to the response object to enable pricing configs?

Happy to take a look! Could you point me to an example where this is done today? https://github.com/search?q=repo%3APortkey-AI%2Fgateway+credits_used&type=code yields no results.

Tripo3D doesn't really have a notion of `credits_used` especially given the async nature of the API where `tasks` are created and then status checked later by clients. It also isn't really feasible to do something like:
```
T0 - take snapshot of balance
T1 - do something
T2 - take snapshot of balance, compute difference
```
because 1) you probably want cost per task and there could be multiple tasks running simultaneously affecting balance 2) multiple requests per API key in between T0 and T2 would also alter balance making it incorrect.

We could try to embed the pricing table into this provider, but it can get out of date quickly and we also have to put in logic as to what is meant by things like `did this call result in "post stylization"`.

https://platform.tripo3d.ai/docs/schema and https://platform.tripo3d.ai/docs/billing

narengogi (changes requested):
Hey @dd-eg-user 
since we do not have a unified signature for these routes,
I'd suggest only keeping the api.ts file and removing the transforms in `src/providers/tripo3d`
this would allow gateway to act as auth layer and allow passthrough proxy requests

so you index file would simply become

```js
import { ProviderConfigs } from '../types';
import Tripo3DAPIConfig from './api';

const Tripo3DConfig: ProviderConfigs = {
  api: Tripo3DAPIConfig,
};

export default Tripo3DConfig;
```

you can remove the changes in `src/providers/types.ts` as well
so finally the only required files are

`src/providers/tripo3d/api.ts`, `src/providers/tripo3d/index.ts`, `globals.ts`

dd-eg-user:
> Hey @dd-eg-user since we do not have a unified signature for these routes, I'd suggest only keeping the api.ts file and removing the transforms in `src/providers/tripo3d` this would allow gateway to act as auth layer and allow passthrough proxy requests
> 
> so you index file would simply become
> 
> ```js
> import { ProviderConfigs } from '../types';
> import Tripo3DAPIConfig from './api';
> 
> const Tripo3DConfig: ProviderConfigs = {
>   api: Tripo3DAPIConfig,
> };
> 
> export default Tripo3DConfig;
> ```
> 
> you can remove the changes in `src/providers/types.ts` as well so finally the only required files are
> 
> `src/providers/tripo3d/api.ts`, `src/providers/tripo3d/index.ts`, `globals.ts`

Thanks for the review. How does this comment and https://github.com/Portkey-AI/gateway/pull/1296#issuecomment-3215531430 work together? The pricing is only relevant on one endpoint. If we remove those endpoints and acts as a passthrough, not sure how we'd add pricing information to the response.

VisargD (approved):

PR #1360, opened 2025-09-23
fix: tools cache_control parameter mapping for bedrock unified messages request transformer

VisargD:
## Type of Change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Unit Tests
- [ ] Integration Tests
- [ ] Manual Testing

## Checklist

### DIFF
--- a/src/providers/bedrock/utils/messagesUtils.ts
+++ b/src/providers/bedrock/utils/messagesUtils.ts
@@ -75,12 +75,14 @@ export const transformToolsConfig = (params: BedrockMessagesParams) => {
             inputSchema: { json: tool.input_schema },
             description: tool.description,
           },
-          ...(tool.cache_control && {
+        });
+        if (tool.cache_control) {
+          tools.push({
             cachePoint: {
               type: 'default',
             },
-          }),
-        });
+          });
+        }
       }
     }
   }


narengogi (approved):

src/providers/bedrock/utils/messagesUtils.ts
```diff
@@ -75,12 +75,14 @@ export const transformToolsConfig = (params: BedrockMessagesParams) => {
             inputSchema: { json: tool.input_schema },
             description: tool.description,
           },
-          ...(tool.cache_control && {
+        });
+        if (tool.cache_control) {
+          tools.push({
             cachePoint: {
               type: 'default',
```
andream16, line 82: Should the value of `type` get customised based on the value passed to `tool.cache_control`'s type?

VisargD: Hi! Gateway currently uses Bedrock Converse APIs internally and that API only supports `default` cachePoint. So there is no other valid mapping possible as of now. Once Bedrock starts supporting some other types, we will add a mapping for them.

Here is Bedrock's current documentation: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent_CachePointBlock.html

andream16 (approved):

PR #1390, opened 2025-10-23
fix: work around `immutable` header guard

### LINKED ISSUE #1389: `TypeError: immutable` with incompatible version of `@hono/node-server`
### What Happened?

When I use a version of `@hono/node-server` greater than `1.14.2`, I get an error response when calling `/v1/chat/completions`:
```json
{"status":"failure","message":"Something went wrong"}
```
In the AI Gateway logs, I see this error:
```txt
tryTargetsRecursively error:  immutable undefined TypeError: immutable
    at appendHeader (node:internal/deps/undici/undici:8659:15)
    at _Headers.append (node:internal/deps/undici/undici:8862:16)
    at Fp.updateHeaders (file:///Users/me/projects/github/gateway/build/start-server.js:2:398498)
    at Fp.create (file:///Users/me/projects/github/gateway/build/start-server.js:2:398112)
```
The behavior may be caused by https://github.com/honojs/node-server/pull/242.

According to the `fetch` [spec](https://fetch.spec.whatwg.org/#headers-guard), response headers with the `immutable` header guard can't be modified, which seems to be what the AI Gateway wants to do in `updateHeaders`.

When I use a lower version of `@hono/node-server`, I get a successful response.

### Relevant Code Snippet

1. `npx @portkey-ai/gateway` (where a version of `@hono/node-server` greater than `1.14.2` gets installed)
2. Call Portkey:
```txt
curl 'http://localhost:8787/v1/chat/completions' \
  -H 'content-type: application/json' \
  -H 'x-portkey-provider: vertex-ai' \
  -H 'x-portkey-vertex-service-account-json: ...' \
  -H 'x-portkey-vertex-project-id: ...' \
  -H 'x-portkey-vertex-region: global' \
  -d '{
    "model": "gemini-2.5-flash",
    "stream": true,
    "messages": [
      {
        "role": "system",
        "content": "You are a helpful assistant"
      },
      {
        "role": "user",
        "content": "what is a portkey?"
      }
    ]
  }'
```

### Your Twitter/LinkedIn

_No response_

mhassan1:
## Description

This PR fixes the `ResponseService.updateHeaders` method to not try to mutate response headers, since it is forbidden by the `fetch` [spec](https://fetch.spec.whatwg.org/#headers-guard).

Resolves https://github.com/Portkey-AI/gateway/issues/1389.

## Motivation

Users who install incompatible (i.e. modern) versions of `@hono/node-server` will end up with response errors when using Portkey.

## Type of Change
- [ ] Documentation update

## How Has This Been Tested?
- [ ] Unit Tests
- [ ] Integration Tests
- [x] Manual Testing

## Checklist

## Related Issues
https://github.com/Portkey-AI/gateway/issues/1389

### DIFF
--- a/src/handlers/services/responseService.ts
+++ b/src/handlers/services/responseService.ts
@@ -61,7 +61,11 @@ export class ResponseService {
       ));
     }
 
-    this.updateHeaders(finalMappedResponse, cache.cacheStatus, retryAttempt);
+    finalMappedResponse = this.updateHeaders(
+      finalMappedResponse,
+      cache.cacheStatus,
+      retryAttempt
+    );
 
     return {
       response: finalMappedResponse,
@@ -99,6 +103,13 @@ export class ResponseService {
     cacheStatus: string | undefined,
     retryAttempt: number
   ) {
+    // Clone response and headers to work around `immutable` header guard
+    response = new Response(response.body, {
+      status: response.status,
+      statusText: response.statusText,
+      headers: new Headers(response.headers),
+    });
+
     // Append headers directly
     response.headers.append(
       RESPONSE_HEADER_KEYS.LAST_USED_OPTION_INDEX,
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
       "dependencies": {
         "@aws-crypto/sha256-js": "^5.2.0",
         "@cfworker/json-schema": "^4.0.3",
-        "@hono/node-server": "^1.3.3",
+        "@hono/node-server": "^1.19.5",
         "@hono/node-ws": "^1.2.0",
         "@portkey-ai/mustache": "^2.1.3",
         "@smithy/signature-v4": "^2.1.1",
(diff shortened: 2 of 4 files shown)


src/handlers/services/responseService.ts
```diff
@@ -99,6 +103,13 @@ export class ResponseService {
     cacheStatus: string | undefined,
     retryAttempt: number
   ) {
+    // Clone response and headers to work around `immutable` header guard
```
narengogi, line 106: expensive operation, have to see if there are better ways to solve this, dont want to. be slowing down the fastest gateway xd

mhassan1: understood. i attempted to access the header guard to modify it in place, but `undici` hides it in a private field ([link](https://github.com/nodejs/undici/blob/b958ea069b9a764910a7f7461acc0f0f660c91ba/lib/web/fetch/headers.js#L633-L639)) and deletes the getters and setters ([link](https://github.com/nodejs/undici/blob/b958ea069b9a764910a7f7461acc0f0f660c91ba/lib/web/fetch/headers.js#L658-L659)).

mhassan1: this should work:
```js
// Clone headers to work around `immutable` header guard
const kHeaders = Object.getOwnPropertySymbols(response).find((symbol) => symbol.toString() === 'Symbol(headers)');
Object.defineProperty(response, kHeaders, {
  value: new Headers(response.headers),
  writable: true,
  enumerable: true,
  configurable: true,
});
```
do we like that better? the benefit is that we don't need to construct a new response, but it is making some unsafe assumptions about internals.

mhassan1: another option:
```js
// Capture internal `getHeadersGuard` and `setHeadersGuard`
let getHeadersGuard, setHeadersGuard
const deleteProperty = Reflect.deleteProperty
Reflect.deleteProperty = function (...args) {
  if (args[1] === 'getHeadersGuard') getHeadersGuard = args[0].getHeadersGuard
  if (args[1] === 'setHeadersGuard') setHeadersGuard = args[0].setHeadersGuard
  return deleteProperty.call(this, ...args)
}

// then
setHeadersGuard(response.headers, 'response')
```

PR #1397, opened 2025-10-28
anthropic cache creation token mapping

narengogi:


roh26it:
@narengogi Are we handling pricing for this in winky?

src/providers/google-vertex-ai/chatComplete.ts
```diff
@@ -772,7 +772,15 @@ export const VertexAnthropicChatCompleteResponseTransform: (
...
-    const { input_tokens = 0, output_tokens = 0 } = response?.usage ?? {};
+    const {
+      input_tokens = 0,
+      output_tokens = 0,
+      cache_creation_input_tokens,
+      cache_read_input_tokens,
+    } = response?.usage ?? {};
+
+    const shouldSendCacheUsage =
+      cache_creation_input_tokens || cache_read_input_tokens;
```
roh26it, line 783: If both are zero, this would be false. Is that intended?

narengogi:
closing as stale

PR #1407, opened 2025-11-03
send hook results when streaming response

narengogi:
I'm not a big fan of the approach here either, Separating hooks execution and response transforms was a good enough abstraction but there is no way to let the stream transformer be made aware of the hooks result than to pass it to the stream transform function

How to test:
configure a beforeRequestHook and set strict-open-ai-compliance flag to false
example payload
```sh
--header 'x-portkey-provider: openai' \
--header 'authorization: sk-' \
--header 'x-portkey-config: {"beforeRequestHooks":[{"type":"guardrail","id":"my_solid_guardrail","checks":[{"id":"default.regexMatch","parameters":{"rule": "*"}}]}],"retry": {"attempts": 1}}' \
--header 'x-portkey-strict-open-ai-compliance: false' \
--data '{
    "model": "gpt-3.5-turbo",
    "max_tokens": 100,
    "stream": true,
    "messages": [
        {
            "role": "system",
            "content": "You are a helpful aaassistant"
        },
        {
            "role": "user",
            "content": "I am drowning under the water"
        }
    ]
}'
```

### DIFF
--- a/src/handlers/responseHandlers.ts
+++ b/src/handlers/responseHandlers.ts
@@ -263,9 +263,6 @@ export async function afterRequestHookHandler(
 
     if (!responseJSON) {
       // For streaming responses, check if beforeRequestHooks failed without deny enabled.
-      const additionalHeaders = {
-        'x-portkey-hook-results': JSON.stringify(hooksResult),
-      };
       if (
         (failedBeforeRequestHooks.length || failedAfterRequestHooks.length) &&
         response.status === 200
@@ -275,12 +272,10 @@ export async function afterRequestHookHandler(
           ...response,
           status: 246,
           statusText: 'Hooks failed',
-          headers: { ...response.headers, ...additionalHeaders },
         });
       }
       return new Response(response.body, {
         ...response,
-        headers: { ...response.headers, ...additionalHeaders },
       });
     }
 
--- a/src/handlers/responseHandlers.ts
+++ b/src/handlers/responseHandlers.ts
@@ -274,9 +274,7 @@ export async function afterRequestHookHandler(
           statusText: 'Hooks failed',
         });
       }
-      return new Response(response.body, {
-        ...response,
-      });
+      return response;
     }
 
     if (shouldDeny) {
--- a/src/handlers/responseHandlers.ts
+++ b/src/handlers/responseHandlers.ts
@@ -99,23 +99,21 @@ export async function responseHandler(
     responseTransformerFunction = undefined;
   }
 
-  if (
-    streamingMode &&
-    isSuccessStatusCode &&
-    isCacheHit &&
-    responseTransformerFunction
-  ) {
-    const streamingResponse = await handleJSONToStreamResponse(
-      response,
-      provider,
-      responseTransformerFunction
-    );
-    return { response: streamingResponse, responseJson: null };
-  }
   if (streamingMode && isSuccessStatusCode) {
     const hooksManager = c.get('hooksManager');
     const span = hooksManager.getSpan(hookSpanId) as HookSpan;
     const hooksResult = span.getHooksResult();
+    if (isCacheHit && responseTransformerFunction) {
+      const streamingResponse = await handleJSONToStreamResponse(
+        response,
+        provider,
+        responseTransformerFunction,
+        strictOpenAiCompliance,
+        responseTransformer as endpointStrings,
+        hooksResult
+      );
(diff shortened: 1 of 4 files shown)


src/handlers/responseHandlers.ts
```diff
@@ -264,7 +272,6 @@ export async function afterRequestHookHandler(
           ...response,
           status: 246,
           statusText: 'Hooks failed',
-          headers: response.headers,
```
roh26it, line 267: Why are we removing this?

src/handlers/responseHandlers.ts
```diff
@@ -110,14 +113,19 @@ export async function responseHandler(
     return { response: streamingResponse, responseJson: null };
   }
   if (streamingMode && isSuccessStatusCode) {
+    const hooksManager = c.get('hooksManager');
```
roh26it, line 103: Should we handle potential null checks here?

If:
- hooksManager is not set in context
- hookSpanId is undefined/null
- The span doesn't exist

This might throw a runtime error and crash the streaming response.

src/handlers/responseHandlers.ts
roh26it: ok, this might be redundant. Good point.

roh26it:
When isCacheHit is true, the code returns early from handleJSONToStreamResponse without including hook results. Is this intentional?

narengogi:
> When isCacheHit is true, the code returns early from handleJSONToStreamResponse without including hook results. Is this intentional?

This has to be handled on our part, concatenating the stream response and then sending it back

roh26it:
> Ok, will you add support for that here?

narengogi:
> Ok, will you add support for that here?

sorry I edited your reply by mistake yes, I will, I was hoping to push it in a subsequent release but let me check if it is a quick fix

narengogi:
@roh26it I've added handling for cached responses to send the hook results and I've tested with the openai sdk to make sure nothing breaks

roh26it (changes requested):

src/handlers/responseHandlers.ts
narengogi: yes this is redundant

src/handlers/responseHandlers.ts
narengogi: hooksManager is set in context for all routes, so not required

VisargD:
- Concatenation logic in middleware for logging
- Should we use a separate flag instead of strict openai compliance since its enabled by default by Portkey SDKs

src/handlers/streamHandler.ts
```diff
@@ -337,6 +347,12 @@ export function handleStreamingMode(
   } else {
     (async () => {
       try {
+        if (!strictOpenAiCompliance) {
+          const hookResultChunk = constructHookResultChunk(hooksResult, fn);
+          if (hookResultChunk) {
```
VisargD, line 361: Won't this always be true for chatComplete, embed, complete and messages endpoints since the return value always be truthy even when there is no hook results

src/handlers/streamHandler.ts
```diff
@@ -434,3 +465,23 @@ export async function handleJSONToStreamResponse(
     statusText: response.statusText,
   });
 }
+
+const constructHookResultChunk = (
+  hooksResult: HookSpan['hooksResult'],
+  fn: endpointStrings
+) => {
+  if (fn === 'chatComplete' || fn === 'complete' || fn === 'embed') {
```
VisargD: I think its better to not maintain another source of truth for hook results eligible endpoint. Hooks middleware already does it.

For now, we can only make an exception for messages since it needs the event_type. But for other, if hooks results is present, we can use the standard structure.

Can you please just verify some other routes to confirm that they won't have the hook_results in the final response?

src/handlers/streamHandler.ts
narengogi: good catch, will check for before_request_hooks.length

src/handlers/streamHandler.ts
narengogi: verified with combination of strict-open-ai-compliance flag, before request hooks and routes

VisargD (approved):

src/handlers/responseHandlers.ts
narengogi: This was infact not redundant :'
https://github.com/Portkey-AI/gateway/pull/1447

PR #1424, opened 2025-11-13
send empty usage object in message start chunk for bedrock messages

### LINKED ISSUE #1417: Bedrock Provider is not fully compatible with Anthropic SDK, docs are outdated.
### What Happened?

I want to use Amazon Bedrock to serve Anthropic Models and also use Portkey as a gateway.
 
I copied the [example code](https://portkey.ai/docs/integrations/llms/bedrock/aws-bedrock#anthropic-typescript-sdk) of using the Anthropic SDK with Portkey's `/messages` endpoint. Note that using this SDK is highly preferable to using the Portkey SDK directly as otherwise there's typing differences between chat completions and Anthropic's message format that must be converted manually.

The following example was run with `bun`, but can easily be run with `node` by switching `Bun.env` for `process.env`.

```
import Anthropic from '@anthropic-ai/sdk';

const anthropic = new Anthropic({
    apiKey: 'dummy', // we will use portkey's provider slug
    baseURL: "https://api.portkey.ai",
    defaultHeaders: { "x-portkey-api-key": Bun.env.PORTKEY_API_KEY },
    fetch: async (input, init) => {
        console.log("fetch", input, init);
        return fetch(input, init);
    }
});

const msg = await anthropic.messages.stream({
    model: "@aws-bedrock-dev/us.anthropic.claude-sonnet-4-20250514-v1:0",
    max_tokens: 1024,
    messages: [{ role: "user", content: "Hello, Claude" }]
});
for await (const message of msg) {
    console.log(JSON.stringify(message, null, 2));
}
```

I received the following validation errors:
```

36 |             }
37 |             if (error instanceof AnthropicError) {
38 |                 return this._emit('error', error);
39 |             }
40 |             if (error instanceof Error) {
41 |                 const anthropicError = new AnthropicError(error.message);
                                            ^
error: undefined is not an object (evaluating 'snapshot.usage.output_tokens = event.usage.output_tokens')
414 |             case 'message_stop':
415 |                 return snapshot;
416 |             case 'message_delta':
417 |                 snapshot.stop_reason = event.delta.stop_reason;
418 |                 snapshot.stop_sequence = event.delta.stop_sequence;
419 |                 snapshot.usage.output_tokens = event.usage.output_tokens;
                               ^
TypeError: undefined is not an object (evaluating 'snapshot.usage.output_tokens = event.usage.output_tokens')
```
It seems that Bedrock does not emit certain usage fields that are required by the Anthropic SDK. Note that we cannot use the AnthropicBedrock SDK with Portkey's /messages endpoint either as the SDK constructs an entirely separate path on top of the base url with the format `/models/{model_name}/{method}` rather than using a unified `/messages` endpoint and passing the model in the request body.

---
Furthermore, the docs currently suggest setting the baseURL for the Anthropic client to `https://api.portkey.ai/v1`, this should actually be `https://api.portkey.ai`. The Anthropic SDK appends /v1/message on top of the baseURL.

Portkey should normalize the stream results from Bedrock so that they have a usage item and don't trigger validation errors from Anthropic's official SDK

OR

Portkey should support the `/models/{model_name}/{method}` format so that the Anthropic Bedrock SDK is natively supported.

---
Docs example should be updated to the following:
```
import Anthropic from '@anthropic-ai/sdk';

   const anthropic = new Anthropic({
     apiKey: 'dummy', // we will use portkey's provider slug
     baseURL: "https://api.portkey.ai",
     defaultHeaders: { "x-portkey-api-key": "YOUR_PORTKEY_API_KEY" }
   });

   const msg = await anthropic.messages.create({
     model: "@your-provider-slug/your-model-name",
     max_tokens: 1024,
     messages: [{ role: "user", content: "Hello, Claude" }],
   });
   console.log(msg);
```

---
Docs example should be updated to mention that Bedrock streaming doesn't work with the `/messages` endpoint right now

### Relevant Code Snippet

_No response_

### Your Twitter/LinkedIn

linkedin.com/in/dayvidwang

narengogi:
this fixes the error reported in #1417 

I've verified this with anthropic sdk

### DIFF
--- a/src/providers/bedrock/messages.ts
+++ b/src/providers/bedrock/messages.ts
@@ -623,7 +623,5 @@ function getMessageStartEvent(fallbackId: string, gatewayRequest: Params) {
   );
   messageStartEvent.message.id = fallbackId;
   messageStartEvent.message.model = gatewayRequest.model as string;
-  // bedrock does not send usage in the beginning of the stream
-  delete messageStartEvent.message.usage;
   return `event: message_start\ndata: ${JSON.stringify(messageStartEvent)}\n\n`;
 }


narengogi:
@dayvidwang for your review

VisargD (approved):

PR #1458, opened 2025-11-27
fix: include thoughtsTokenCount in completion_tokens for Gemini

voodoohop:
- `completion_tokens` now includes `thoughtsTokenCount` for Gemini thinking models
- Fixes both streaming and non-streaming responses
- Aligns with OpenAI format where `completion_tokens` is the total (reasoning + output)
- `completion_tokens_details.reasoning_tokens` provides the breakdown

**Problem:**
- Gemini returns `candidatesTokenCount` (output) and `thoughtsTokenCount` (reasoning) separately
- OpenAI format expects `completion_tokens` to be the sum of both
- Previously only `candidatesTokenCount` was used

**Changes:**
- `src/providers/google/chatComplete.ts`
- `src/providers/google-vertex-ai/chatComplete.ts`

### DIFF
--- a/src/providers/google-vertex-ai/chatComplete.ts
+++ b/src/providers/google-vertex-ai/chatComplete.ts
@@ -527,7 +527,7 @@ export const GoogleChatCompleteResponseTransform: (
         }) ?? [],
       usage: {
         prompt_tokens: promptTokenCount,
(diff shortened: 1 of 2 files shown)


siddharthsambharia-portkey:
@narengogi can you look at this?

PR #1462, opened 2025-12-04
Update gemini response transform: calculate completion_tokens as the sum of candidatesTokenCount and thoughtsTokenCount to match OpenAI API expectations

eulervoid:
**Description:**
With the OpenAI API, `completion_tokens` is meant to be the total number of completions tokens, including the ones further specified in `completion_tokens_details`, like `reasoning_tokens`. The current behavior of the `google` provider in portkey is to return them separately, which leads to incorrect usage calculations down the line.

This PR tries to fix this by:
- Calculating `completion_tokens` as the sum of `candidatesTokenCount` and `thoughtsTokenCount` to match the convention of the OpenAI API. I think this should probably also apply to `audio_tokens`, but i'm not as confident there, which is why i'm focusing on reasoning first with this PR.
- Adding this update for both `google` and `google-vertex-ai` providers.

**Tests Run/Test cases added:**
- No additional tests added
- Unfortunately, i wasn't able to run all the tests due to missing credentials

**Type of Change:**

- [ ] Documentation update

### DIFF
--- a/src/providers/google-vertex-ai/chatComplete.ts
+++ b/src/providers/google-vertex-ai/chatComplete.ts
@@ -527,7 +527,7 @@ export const GoogleChatCompleteResponseTransform: (
         }) ?? [],
       usage: {
         prompt_tokens: promptTokenCount,
-        completion_tokens: candidatesTokenCount,
+        completion_tokens: candidatesTokenCount + thoughtsTokenCount,
         total_tokens: totalTokenCount,
         completion_tokens_details: {
           reasoning_tokens: thoughtsTokenCount,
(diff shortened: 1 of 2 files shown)


narengogi:
valid @eulervoid reviewing

narengogi:
with vertex, I just want to verify if the behaviour is same as gemini for anthropic models on vertex also

narengogi (commented):
Looks good to me, let me update this PR with the relevant changes for anthropic on vertex as well

VisargD:
@narengogi - Should we wait for your updates?

narengogi:
fixing

metinogurlu:
Hi thanks for the PR. Waiting as well for this change, any ETA?

narengogi (approved):

PR #1475, opened 2025-12-17
execute guardrails sequentially

narengogi:
This allows users to execute guardrail checks sequentially, 
useful for example when you have two regex replace checks in the same guardrails.

Previously, only the second replace guardrail would take effect, by passing this flag, the mutated response from one check is passed to the next one in the order they are passed in the config


VisargD (approved):

VisargD:
TODO: We should add sequential flag on hooks level as well just like we did for checks

PR #1492, opened 2025-12-30
[bug/behaviour] only execute enabled checks when sequential guardrails are enabled

narengogi:


src/middlewares/hooks/index.ts
```diff
@@ -365,26 +365,28 @@ export class HooksManager {
...
-            result.transformedData &&
-            (result.transformedData.response.json ||
-              result.transformedData.request.json)
-          ) {
-            span.setContextAfterTransform(
-              result.transformedData.response.json,
-              result.transformedData.request.json
+        hook.checks
+          .filter((check: Check) => check.is_enabled !== false)
+          .forEach(async (check: Check) => {
```
VisargD: forEach with async/await might not give consistent result for set context.

narengogi: updated with a for loop

PR #1504, opened 2026-01-15
Initial Crowdstrike AIDR Guardrail Plugin

ryanmeans:
**Description:** (required)
This is effectively an update for the Pangea Guardrails plugin. First, it is a rebrand, since Pangea has been acquired by crowdstrike, and has a new product name

It's updated for much newer APIs. The current APIs support blocking and redaction based on configured upstream rules. The new endpoint is designed to work with structued LLM inputs, e.g. you should pass it the entire `messages` array, but also supports having `tools` and other properties a well. 

I think we should consider removing the Pangea guardrails as well. 

- [x] Added tests for the new guardrails

**Type of Change:**

- [ ] Documentation update

### DIFF
--- a/plugins/crowdstrike-aidr/guardChatCompletion.ts
+++ b/plugins/crowdstrike-aidr/guardChatCompletion.ts
@@ -1,0 +1,151 @@
+import {
+  HookEventType,
+  PluginContext,
+  PluginHandler,
+  PluginParameters,
+} from '../types';
+import { post, HttpError } from '../utils';
+import { VERSION } from './version';
+
+export const handler: PluginHandler = async (
+  context: PluginContext,
+  parameters: PluginParameters,
+  eventType: HookEventType
+) => {
+  let error = null;
+  let verdict = true;
+  let data = null;
+
+  if (!parameters.credentials?.baseUrl) {
+    return {
+      error: `'parameters.credentials.baseUrl' must be set`,
+      verdict: true,
+      data,
+    };
+  }
+
+  if (!parameters.credentials?.apiKey) {
+    return {
+      error: `'parameters.credentials.apiKey' must be set`,
+      verdict: true,
+      data,
+    };
+  }
+
+  const url = `${parameters.credentials.baseUrl}/v1/guard_chat_completions`;
+  const target = eventType === 'beforeRequestHook' ? 'request' : 'response';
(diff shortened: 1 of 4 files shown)


plugins/crowdstrike-aidr/guardChatCompletion.ts
```diff
@@ -0,0 +1,151 @@
...
+  const url = `${parameters.credentials.baseUrl}/v1/guard_chat_completions`;
+  const target = eventType === 'beforeRequestHook' ? 'request' : 'response';
+  const json = context[target].json;
+  const aidrEventType = target === 'request' ? 'input' : 'output';
+
+  const requestBody: object = {
+    guard_input: json,
+    event_type: aidrEventType,
+    app_id: 'Portkey AI Gateway',
+    // TODO: Add as much other metadata as we have
```
ryanmeans, line 44: Is there anything else present in the `context` object which I can add here to decorate the request? e.g. a source IP address or some kind of user identifiers.

plugins/crowdstrike-aidr/guardChatCompletion.ts
```diff
@@ -0,0 +1,151 @@
...
+    return {
+      error: `'parameters.credentials.apiKey' must be set`,
+      verdict: true,
+      data,
+    };
+  }
+
+  const url = `${parameters.credentials.baseUrl}/v1/guard_chat_completions`;
+  const target = eventType === 'beforeRequestHook' ? 'request' : 'response';
+  const json = context[target].json;
```
VisargD, line 37: Since the whole request body is being passed without any transformation, this would only work for chat completions API but it might fail for OpenAI responses API and Anthropic messages API format if the Crowdstrike guardrails API does not support them.

Are responses and messages format supported? I do not see them in the documentation

VisargD: Marking this as a known limitation and merging this for now.

ryanmeans: @VisargD  Thanks, I'll look into it on our end. Right now the endpoint supports any kind of JSON, but it does implicitly expect the format to be the OpenAI completions AI format and we do have special logic for the messages array which might be lost.

PR #1485, opened 2025-12-23
provider: Databricks

azaccor:
**Description:** (required)
- Add support for Databricks AI Gateway as a Portkey provider
- Implements chat completions, text completions, and embeddings
- Uses unified /invocations endpoint for all operations
- Maintains OpenAI-compatible request/response format
- Supports workspace-based and custom base URLs
- Bearer token authentication with Databricks PATs

**Tests Run/Test cases added:** (required)
- [X] Tested with Llama 4 Maverick, Claude Sonnet 4.5, and GTE embedding models to ensure it worked across multiple functionalities, not just completion or chat, and also that the /invocations endpoint standard on Databricks model serving was able to successfully exception handle non OpenAI API style requests

**Type of Change:**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)

PR #1505, opened 2026-01-19
Feat/databricks provider

b4s36t4:
**Description:** (required)
- continuation of #1485

- [ ] Description of test case

**Type of Change:**

- [ ] Documentation update


narengogi (dismissed):

src/globals.ts
```diff
@@ -113,6 +113,7 @@ export const ORACLE: string = 'oracle';
 export const IO_INTELLIGENCE: string = 'iointelligence';
 export const AIBADGR: string = 'aibadgr';
 export const OVHCLOUD: string = 'ovhcloud';
+export const DATABRICKS_AI: string = 'databricks-ai';
```
azaccor, line 116: I registered the provider as 'databricks' in providers/index.ts. I have a relatively strong preference that the provider identifier be just 'databricks' to match the pattern of other major providers like 'openai', 'anthropic', etc. Can we keep it as 'databricks' instead of 'databricks-ai'?

azaccor:
Hello @b4s36t4 and @narengogi , thanks so much for your reviews! 
This makes sense to me to remove the baseURL and only keep the workspace based URLs. My only question is on the naming convention if we're able to keep it as 'databricks' rather than 'databricks-ai'? 
Thanks again!

azaccor:
Hi all, following up on the above. Anything I can do to expedite the review? 
Thanks so much!

azaccor:
Hi all, circling back here. Anything I can do to help with the review? Would love to get this merged if possible. 
Cheers

narengogi (approved):

PR #1521, opened 2026-02-15
Add support for Vertex AI gemini and normal gemini count token

LauHoiYanGladys:
**Description:** (required)
- Add support for Vertex AI gemini and normal gemini count token

- [x] Tested via manual queries to those two services run locally

**Type of Change:**

- [ ] Documentation update

### DIFF
--- a/src/providers/google/countTokens.ts
+++ b/src/providers/google/countTokens.ts
@@ -1,0 +1,49 @@
+import { GOOGLE } from '../../globals';
+import { ErrorResponse, ProviderConfig } from '../types';
+import { generateInvalidProviderResponseError } from '../utils';
+import {
+  GoogleChatCompleteConfig,
+  GoogleErrorResponse,
+  GoogleErrorResponseTransform,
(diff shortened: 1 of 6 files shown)


src/providers/google/countTokens.ts
```diff
@@ -0,0 +1,49 @@
...
+};
+
+interface GoogleCountTokensResponse {
+  totalTokens: number;
+  cachedContentTokenCount?: number;
+}
+
+// GoogleCountTokensResponseTransform maps Gemini's { totalTokens } to the
+// gateway's unified { input_tokens } format.
+export const GoogleCountTokensResponseTransform: (
```
narengogi, line 31: @LauHoiYanGladys 
can you confirm if the  response from gemini count tokens follows the same structure as the anthropic messages count tokens endpoint?
if not please handle it here.
the messages count tokens endpoint is used by claude code for validating the token counts and the repsonse structure has to be the same

PR #1524, opened 2026-02-15
improvement: Adds a user-agent header for Anthropic API calls

mikelambert:
**Description:** (required)
- Adds a user-agent header for Anthropic API calls

Tests that the anthropic api passes the user-agent header and the api key

**Type of Change:**
- [X] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

### DIFF
--- a/src/providers/anthropic/api.ts
+++ b/src/providers/anthropic/api.ts
@@ -1,4 +1,5 @@
 import { ProviderAPIConfig } from '../types';
+import { version as packageVersion } from '../../../package.json';
 
 const AnthropicAPIConfig: ProviderAPIConfig = {
   getBaseURL: () => 'https://api.anthropic.com/v1',
@@ -8,6 +9,7 @@ const AnthropicAPIConfig: ProviderAPIConfig = {
       providerOptions.apiKey || providerOptions.anthropicApiKey || '';
     const headers: Record<string, string> = {
       'X-API-Key': apiKey,
+      'User-Agent': `portkey-ai-gateway/${packageVersion}`,
     };
 
     // Accept anthropic_beta and anthropic_version in body to support enviroments which cannot send it in headers.
--- a/src/providers/anthropic/api.test.ts
+++ b/src/providers/anthropic/api.test.ts
@@ -1,0 +1,34 @@
+import AnthropicAPIConfig from './api';
+import { version as packageVersion } from '../../../package.json';
+
+describe('AnthropicAPIConfig', () => {
+  describe('headers', () => {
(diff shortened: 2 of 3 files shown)


src/providers/anthropic/api.ts
```diff
@@ -8,6 +9,7 @@ const AnthropicAPIConfig: ProviderAPIConfig = {
       providerOptions.apiKey || providerOptions.anthropicApiKey || '';
     const headers: Record<string, string> = {
       'X-API-Key': apiKey,
+      'User-Agent': `portkey-ai-gateway/${packageVersion}`,
```
narengogi, line 12: would recommend avoiding any identifiable information currently much like how browsers scramble different identifiers to avoid detection because we dont want anthropic blocking people arbitrarily and slowing down their requests

narengogi:
@mikelambert I can see you work at Anthropic, any specific reason for this PR?

mikelambert:
Heya yup! I work at Anthropic, wasn't trying to pull one over on you :)

Context: Our datascience team would love to know what libraries people are _actually_ using to access our API services, but our privacy terms make it very difficult to actually look at any prompt keywords like that, so we're not able to do any identification of common libraries across API keys/orgs/users/etc. I volunteered to go make this kind of change across a bunch of OSS libraries...some already were passing user-agents, some were not yet doing so (ie langchain, portkey, etc).

> would recommend avoiding any identifiable information currently much like how browsers scramble different identifiers to avoid detection because we dont want anthropic blocking people arbitrarily and slowing down their requests

I assume this portkey library is intended for individuals/companies to use with their own API keys, and such "build against a library which then swaps out backend providers in response to X" is totally valid use case used by many of our customers, so I don't see any reason for us to want to block/slow requests...

Is your concern on identification that we'd act against portkey library users for some reason? I don't see anything in the README that seems to imply anything sketchy going on with portkey that'd require obfuscation, but maybe I'm not understanding what you're worried about! If there's something that I can help to clarify / de-confuse, I'm happy to help...but if you ultimately end up feeling uncomfortable to add this, I understand, though personally would love to better understand your concern!

narengogi:
@mikelambert 
cool I'll approve it,
my concern was if anthropic would block requests from certain downstream clients 
(I suppose you have a million other ways to do block requests from identifiers anyways if not the agent header)

narengogi (approved):

mikelambert:
Thank you narengogi! Do you want me to jump through the GPG signing here, or does that get fixed as part of squashing to master and signing on your/github end?

narengogi:
would be nice if you can sign but I'll merge this

PR #1525, opened 2026-02-16
feat: Integration of Akto plugin

rahul-akto:
## Description
This PR integrates the Akto plugin into the Portkey AI Gateway. It enables security scanning for AI applications by connecting the gateway to Akto's guardrails, providing comprehensive validation of requests and responses to/from LLM models.


rahul-akto:
As we are going to raise from master

PR #1534, opened 2026-02-21
fix: return tool_calls finish_reason for Gemini function calls

roh26it:
## Summary

- Fixes Gemini/Vertex AI returning `finish_reason: "stop"` instead of `"tool_calls"` when function calls are present
- Gemini API returns `finishReason: "STOP"` even when the response contains function calls (there is no `FUNCTION_CALL` finish reason in Gemini's API)
- This fix detects tool calls in the response content and returns `finish_reason: "tool_calls"` for OpenAI compatibility

## Changes

- Modified `GoogleChatCompleteResponseTransform` (non-streaming) for both Google and Vertex AI
- Modified `GoogleChatCompleteStreamChunkTransform` (streaming) for both Google and Vertex AI

## Test plan

- [ ] Verify Gemini tool calls return `finish_reason: "tool_calls"` in non-streaming mode
- [ ] Verify Gemini tool calls return `finish_reason: "tool_calls"` in streaming mode
- [ ] Verify regular completions still return appropriate finish reasons

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/providers/google-vertex-ai/chatComplete.ts
+++ b/src/providers/google-vertex-ai/chatComplete.ts
@@ -593,10 +593,13 @@ export const GoogleChatCompleteResponseTransform: (
           return {
             message: message,
             index: index,
-            finish_reason: transformFinishReason(
-              generation.finishReason as GOOGLE_GENERATE_CONTENT_FINISH_REASON,
-              strictOpenAiCompliance
-            ),
+            finish_reason:
+              toolCalls.length > 0
+                ? 'tool_calls'
+                : transformFinishReason(
+                    generation.finishReason as GOOGLE_GENERATE_CONTENT_FINISH_REASON,
+                    strictOpenAiCompliance
+                  ),
             logprobs,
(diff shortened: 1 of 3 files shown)


narengogi (changes requested):
would recommend extracting this to `google-vertex-ai/utils.ts` and also making use of `FINISH_REASON.tool_calls` instead of hard coding the string

src/providers/google-vertex-ai/chatComplete.ts
```diff
@@ -732,12 +737,15 @@ export const GoogleChatCompleteStreamChunkTransform: (
     provider: GOOGLE_VERTEX_AI,
     choices:
       parsedChunk.candidates?.map((generation, index) => {
+        const containsToolCalls = hasToolCalls(generation.content?.parts);
```
narengogi, line 740: you'd want to store this variable in streamState and make use of it because the last chunk might not have the tool call in the chunk

narengogi (changes requested):

PR #1593, opened 2026-04-07
fix(anthropic): pass through top-level cache_control for automatic caching

bhuvansingla:
## Summary

- Anthropic supports [automatic prompt caching](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#automatic-prompt-caching) via a top-level `cache_control` field on the request body, which automatically applies the cache breakpoint to the last cacheable block
- This field was not present in `AnthropicChatCompleteConfig`, so it was silently dropped during OpenAI→Anthropic request transformation
- Added `cache_control` as a passthrough parameter in the config

### DIFF
--- a/src/providers/anthropic/chatComplete.ts
+++ b/src/providers/anthropic/chatComplete.ts
@@ -469,6 +469,10 @@ export const AnthropicChatCompleteConfig: ProviderConfig = {
     param: 'thinking',
     required: false,
   },
+  cache_control: {
+    param: 'cache_control',
+    required: false,
+  },
 };
 
 interface AnthorpicTextContentItem {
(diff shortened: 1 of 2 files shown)


src/providers/anthropic/chatComplete.ts
```diff
@@ -469,6 +469,10 @@ export const AnthropicChatCompleteConfig: ProviderConfig = {
     param: 'thinking',
     required: false,
   },
+  cache_control: {
```
whysumedh, line 472: hey can you also add this to src/providers/anthropic-base/messages.ts

whysumedh (changes requested):

src/providers/anthropic/chatComplete.ts
bhuvansingla: @whysumedh Done: https://github.com/Portkey-AI/gateway/pull/1593/commits/0c911aa023932963b86a6dcd82a0f32621eea696

whysumedh (approved):

narengogi (approved):

bhuvansingla:
Closing as it is already supported now.

PR #1615, opened 2026-04-19
fix(mistral-ai): forward response_format for JSON schema

### LINKED ISSUE #1559: Update mistral integration
mistral integration is missing json schema field mapping, needs to be fixed

arun2728:
**Description:** (required)
- Register `response_format` on the Mistral AI chat completion `ProviderConfig` so the gateway forwards OpenAI-compatible `response_format` (including `type: "json_schema"` and nested `json_schema`) to Mistral’s `/v1/chat/completions` API instead of dropping it during `transformUsingProviderConfig`.
- When `response_format.type` is `json_schema` and the inner `json_schema` object omits `name`, set a default `name` of `"response"` so payloads match Mistral’s OpenAPI (required `name` + `schema` on the inner object) while staying compatible with clients that omit `name`.

Fixes #1559

- [x] Ran `transformUsingProviderConfig(MistralAIChatCompleteConfig, …)` with `response_format.type: "json_schema"` and verified the transformed body includes `response_format` and defaulted `json_schema.name` when omitted.
- [x] End-to-end: `curl` to local gateway (`x-portkey-provider: mistral-ai`) with `response_format` / JSON schema and confirmed a successful structured response from Mistral.

**Type of Change:**

- [ ] Documentation update

### DIFF
--- a/src/providers/mistral-ai/chatComplete.ts
+++ b/src/providers/mistral-ai/chatComplete.ts
@@ -100,6 +100,30 @@ export const MistralAIChatCompleteConfig: ProviderConfig = {
     param: 'parallel_tool_calls',
     default: null,
   },
+  response_format: {
+    param: 'response_format',
+    default: null,
+    transform: (params: Params) => {
+      const rf = params.response_format;
+      if (!rf) {
+        return null;
+      }
+      if (rf.type !== 'json_schema' || !rf.json_schema) {
+        return rf;
+      }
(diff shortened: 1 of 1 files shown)


arun2728:
Hi @roh26it, just following up on the PR I raised last week. The fix is non-breaking and only addresses the issue without affecting existing functionality. Please take a look whenever you get a chance. Thanks!

PR #1624, opened 2026-04-27
feat(promptsecurity): extend guardrail plugin with full Prompt Security API surface

naresh4dev:
**Description:** (required)
- Extended the Prompt Security plugin to support all 15 prompt-side and 12 response-side detectors via policy passthrough previously the plugin only sent raw text and read `passed` as a boolean.
- Added `getSystemPrompt()` and `buildProtectPayload()` helpers in `shared.ts` to auto-extract system prompts and build the full API request body (`system_prompt`, `policy`, `user`, `user_groups`, `monitor_only`).
- Added redaction support in both `protectPrompt` and `protectResponse` handlers: when the API returns `modified_text` and `redact=true`, the request/response content is rewritten via `setCurrentContentPart()`.
- Added null-safe API response parsing and text fallback to `context.request.text` / `context.response.text` when structured JSON is unavailable.
- Added `parameters` schema to `manifest.json` for both functions (`policy`, `redact`, `monitorOnly`, `user`, `userGroups`).
- Rewrote test suite with 26 test cases (6 unit + 20 integration).

- [ ] Run `npx jest plugins/promptsecurity/promptsecurity.test.ts` with valid `PROMPT_SECURITY_API_DOMAIN` and `PROMPT_SECURITY_API_KEY` env vars
- [ ] Verify all 26 tests pass (6 unit + 20 integration)
- [ ] Verify redaction tests show `modified_text` replacing PII in `transformedData`
- [ ] Confirm missing-credentials tests still fail gracefully
- [ ] Verify `npm run build-plugins` succeeds

**Type of Change:**

- [ ] Documentation update

### DIFF
--- a/plugins/promptsecurity/manifest.json
+++ b/plugins/promptsecurity/manifest.json
@@ -27,9 +27,41 @@
       "description": [
         {
           "type": "subHeading",
-          "text": "Protect a user prompt before it is sent to the LLM."
+          "text": "Protect a user prompt before it is sent to the LLM. Supports configurable policy for detectors such as Prompt Injection, Sensitive Data, Secrets, Harmful Content, Topics, Language, Regex, Code, URLs, Sentiment, Token Limitation, Token Rate Limit, Unicode, Data Privacy Guidelines, and Natural Language Guardrails."
         }
-      ]
+      ],
+      "parameters": {
+        "type": "object",
+        "properties": {
+          "policy": {
+            "type": "object",
+            "label": "Policy",
+            "description": "Prompt Security policy JSON object to configure individual detectors (e.g. Prompt Injection Engine, Sensitive Data, Secrets, Harmful Content Moderator, Topics Detector, Language Detector, Regex, Code Detector, URLs Detector, Sentiment, Token Limitation, Token Rate Limit, Unicode Detector, Data Privacy Guidelines, Natural Language Guardrails)"
+          },
+          "redact": {
+            "type": "boolean",
+            "label": "Enable Redaction",
+            "description": "When enabled, applies text modifications returned by detectors (e.g. PII sanitization, secret redaction, regex redaction) to the request before sending to the LLM",
+            "default": false
+          },
+          "monitorOnly": {
+            "type": "boolean",
+            "label": "Monitor Only",
+            "description": "When enabled, detectors run and log findings but do not block the request",
+            "default": false
+          },
+          "user": {
+            "type": "string",
+            "label": "User",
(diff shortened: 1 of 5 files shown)


plugins/promptsecurity/manifest.json
davidm-ps, line 23: we suggested in [https://github.com/Portkey-AI/gateway/pull/1607/changes](url) to rename this to "Prompt Security Prompt

plugins/promptsecurity/manifest.json
```diff
@@ -27,9 +27,41 @@
...
+          "userGroups": {
+            "type": "array",
+            "label": "User Groups",
+            "description": "User group memberships for group-based policy enforcement"
+          }
+        }
+      }
     },
     {
       "name": "Protect Response",
```
davidm-ps, line 67: we suggested in [https://github.com/Portkey-AI/gateway/pull/1607/changes](url) to rename this to "Prompt Security Response

plugins/promptsecurity/protectPrompt.ts
```diff
@@ -14,18 +18,37 @@ export const handler: PluginHandler = async (
...
+    const payload = buildProtectPayload(text, 'prompt', context, parameters);
+    let response = await promptSecurityProtectApi(
       parameters.credentials,
-      scanPromptObject
+      payload
     );
-    data = data.result.prompt;
-    verdict = data.passed;
+    data = response.result?.prompt ?? null;
+    verdict = data?.passed ?? false;
```
davidm-ps, line 36: we suggest to use
`verdict = data.action !== 'block';`

davidm-ps: that way, if action is `action:"modify"` it will pass, sanitized

davidm-ps: `verdict = data?.action ? data.action !== 'block' : (data?.passed ?? false);`
would have both options

plugins/promptsecurity/protectResponse.ts
```diff
@@ -14,18 +18,41 @@ export const handler: PluginHandler = async (
...
+
+    let response = await promptSecurityProtectApi(
       parameters.credentials,
-      scanResponseObject
+      payload
     );
-    data = data.result.response;
-    verdict = data.passed;
+    data = response.result?.response ?? null;
+    verdict = data?.passed ?? false;
```
davidm-ps, line 40: same as for protectPrompt

PR #1631, opened 2026-05-05
fix(huggingface): use HF inference router for default base URL

### LINKED ISSUE #1626: Hugging Face provider does not work with the documented setup
### What Happened?

I am trying to use Hugging Face models through Portkey.

I followed the Portkey Hugging Face documentation and added Hugging Face as a provider in the Model Catalog. I also manually created the model through the Model Manager.

Reference:
- Portkey Hugging Face docs: https://portkey.ai/docs/integrations/llms/huggingface

```python
import os
from portkey_ai import Portkey
from dotenv import load_dotenv

load_dotenv()

portkey = Portkey(api_key=os.environ.get("PORTKEY_API_KEY"))

response = portkey.chat.completions.create(
    model="@huggingface/meta-llama/Llama-3.2-3B-Instruct",
    messages=[{"role": "user", "content": "Hello!"}]
)

print(response.choices[0].message.content)

```

Image

However, requests through Portkey fail when using the Hugging Face provider.

This does not seem specific to one model. I would like to understand how Hugging Face models in general should be called through Portkey.

```
$ python main.py 
...
...
openai.BadRequestError: Error code: 400 - {'status': 'failure', 'message': 'Invalid provider passed'}
```

The same Hugging Face token works when I call Hugging Face directly, so the token itself seems valid.

One thing I noticed is that the Gateway code appears to use `https://api-inference.huggingface.co` as the default Hugging Face base URL:

- https://github.com/Portkey-AI/gateway/blob/351692fd9236af222168134b416924fae0bdba23/src/providers/huggingface/api.ts#L6

But Hugging Face’s current OpenAI-compatible chat completions documentation uses the router endpoint:

- Hugging Face Inference Providers docs: https://huggingface.co/docs/api-inference

So I am wondering if the Hugging Face provider is still using an older API URL, or if there is a recommended way to configure `huggingfaceBaseUrl` or Custom Host for the Hugging Face provider.

I also noticed that when I try to add a custom model in the Portkey dashboard, the Base Model selector shows `No results found`.I already reported the 403 issue separately in https://github.com/Portkey-AI/gateway/issues/1625, so this issue is mainly about how to use the Hugging Face provider through Portkey.

_No response_

### Relevant Code Snippet

_No response_

### Your Twitter/LinkedIn

ystkfujii

cybertron288:
Closes #1626

## What
Update the default Hugging Face base URL from `https://api-inference.huggingface.co` (the legacy HF Serverless Inference API) to `https://router.huggingface.co` (the new HF Inference Providers OpenAI-compatible router), and drop the per-model path from the default endpoint construction since the router takes the model in the request body.

## Why
The legacy serverless API endpoint with the `/models/{model}/v1/chat/completions` path has been superseded by HF's unified Inference Providers router (per https://huggingface.co/docs/inference-providers). Issue #1626 reports the documented Hugging Face setup no longer working through Portkey, and points at the outdated default URL.

## Notes
- Opened as draft — please review before marking ready.
- The `huggingfaceBaseUrl` override path already used a bare `/v1/chat/completions` endpoint (no `/models/{model}`), so dedicated-endpoint users are unaffected.
- Verification: no test suite was run locally — the change is a config-only edit. Reviewers should sanity-check the new URL and the assumption that the inference router accepts the OpenAI-compatible chat-completions / completions schema with model in the body.

### DIFF
--- a/src/providers/huggingface/api.ts
+++ b/src/providers/huggingface/api.ts
@@ -3,23 +3,18 @@ import { ProviderAPIConfig } from '../types';
 const HuggingfaceAPIConfig: ProviderAPIConfig = {
   getBaseURL: ({ providerOptions }) => {
     return (
-      providerOptions.huggingfaceBaseUrl ||
-      'https://api-inference.huggingface.co'
+      providerOptions.huggingfaceBaseUrl || 'https://router.huggingface.co'
     );
   },
   headers: ({ providerOptions }) => ({
     Authorization: `Bearer ${providerOptions.apiKey}`,
   }),
-  getEndpoint: ({ fn, gatewayRequestBodyJSON, providerOptions }) => {
-    const { model } = gatewayRequestBodyJSON;
-    const modelPath = providerOptions.huggingfaceBaseUrl
-      ? ''
-      : `/models/${model}`;
+  getEndpoint: ({ fn }) => {
     switch (fn) {
       case 'chatComplete':
-        return `${modelPath}/v1/chat/completions`;
+        return '/v1/chat/completions';
       case 'complete':
-        return `${modelPath}/v1/completions`;
+        return '/v1/completions';
       default:
         return '';
     }


ystkfujii:
@cybertron288 

Are you planning to mark the PR as ready?

cybertron288:
Yes, marked as ready for review now. Thanks for the nudge @ystkfujii!

PR #1644, opened 2026-05-12
Create Cato Networks plugin

trexinc:
**Description:** (required)
Add Cato Networks plugin for guardrail functionality

- [ ] Test for the plugin

**Type of Change:**

### DIFF
--- a/plugins/cato/analyze.ts
+++ b/plugins/cato/analyze.ts
@@ -523,11 +523,14 @@ export const handler: PluginHandler = async (
   let transformed = false;
 
   const credentials = parameters.credentials as CatoCredentials | undefined;
-  const apiKey = credentials?.apiKey || (process.env.CATO_API_KEY as string);
+  const apiKey = credentials?.apiKey;
 
   if (!apiKey) {
     return {
-      error: { message: 'Cato API key is required but not configured.' },
+      error: {
+        message:
+          'Cato API key is required but not configured in plugin credentials.',
+      },
       verdict: true,
       data: null,
       transformedData,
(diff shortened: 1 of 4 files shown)


plugins/cato/analyze.ts
```diff
@@ -0,0 +1,676 @@
...
+      data: null,
+      transformedData,
+      transformed,
+    };
+  }
+
+  const apiBase =
+    credentials?.apiBase ||
+    (process.env.CATO_API_BASE as string) ||
+    CATO_DEFAULT_BASE_URL;
```
b4s36t4: @trexinc Can we please remove environment variable references? You can expect these values from Parameters if not found we can throw validation error.

plugins/cato/analyze.ts
```diff
@@ -0,0 +1,676 @@
...
+        status: e.response?.status,
+      };
+    } else {
+      if (e && typeof e === 'object') {
+        delete e.stack;
+      }
+      error = e;
+    }
+    const failOpen = parameters.failOpen !== false;
+    verdict = failOpen;
```
b4s36t4: Can we remove this? Gateway already handles this and Portkey Platform the result as errored and request is also passed-through by design.

plugins/cato/analyze.ts
trexinc: @b4s36t4 fixed

plugins/cato/analyze.ts
trexinc: @b4s36t4 fixed

PR #1646, opened 2026-05-12
fix(gemini): use thinkingBudget for Gemini 2.5, thinkingLevel for 3+

### LINKED ISSUE #1639: `reasoning_effort` silently no-ops for Vertex `gemini-2.5-flash-lite` on `/v1/responses`
### What Happened?

On `POST /v1/responses` with a `vertex-ai` target, the top-level `reasoning_effort` shorthand has no effect on `gemini-2.5-flash-lite`: the model returns zero reasoning tokens for `"low"`, `"medium"`, or `"high"`. The same field works on the same model via `/v1/chat/completions`, and on `/v1/responses` for `gemini-2.5-flash` (non-lite) and `gemini-3-flash-preview`. The OpenAI canonical nested form `reasoning: {effort: ...}` works on `/v1/responses` for flash-lite — that's the workaround.

The request returns HTTP 200 with no error or warning. The bug is also hidden on every other Gemini variant because they default to thinking-on; flash-lite is the only Gemini model with a default-off thinking budget, which is why it's the one place the broken translation is observable.

## Reproduction

`openai-python ≥ 1.66.0` (when `client.responses` was introduced).

```python
import asyncio, json, os
from openai import AsyncOpenAI


def cfg(override_params, vkey, provider="vertex-ai"):
    return {
        "strategy": {"mode": "single"},
        "targets": [{"virtual_key": vkey, "provider": provider, "override_params": override_params}],
    }


async def probe(label, override_params, vkey):
    client = AsyncOpenAI(
        api_key=os.environ["PORTKEY_API_KEY"],
        base_url="https://api.portkey.ai/v1",
        default_headers={"x-portkey-config": json.dumps(cfg(override_params, vkey))},
    )
    try:
        resp = await client.responses.create(
            model="placeholder",  # actual model comes from override_params
            input="Briefly summarize this short message.",
            store=False,
        )
    finally:
        await client.close()
    rt = (
        resp.usage.output_tokens_details.reasoning_tokens
        if (resp.usage and getattr(resp.usage, "output_tokens_details", None))
        else 0
    )
    print(f"  {label}: reasoning_tokens={rt}")


async def main():
    LITE_VKEY = "<your-vertex-ai-virtual-key>"
    FLASH_VKEY = "<your-vertex-ai-virtual-key>"

    # /v1/responses with bare reasoning_effort — silently no-ops on flash-lite
    await probe("flash-lite + reasoning_effort=low (BARE)",
                {"model": "gemini-2.5-flash-lite", "reasoning_effort": "low"}, LITE_VKEY)
    await probe("flash-lite + reasoning_effort=high (BARE)",
                {"model": "gemini-2.5-flash-lite", "reasoning_effort": "high"}, LITE_VKEY)

    # /v1/responses with nested OpenAI-canonical form — works
    await probe("flash-lite + reasoning={effort:low} (NESTED)",
                {"model": "gemini-2.5-flash-lite", "reasoning": {"effort": "low"}}, LITE_VKEY)

    # Controls — bare form works on these
    await probe("CONTROL gemini-3-flash + reasoning_effort=low (BARE)",
                {"model": "gemini-3-flash-preview", "reasoning_effort": "low"}, FLASH_VKEY)
    await probe("CONTROL gemini-2.5-flash + reasoning_effort=low (BARE)",
                {"model": "gemini-2.5-flash", "reasoning_effort": "low"}, LITE_VKEY)


asyncio.run(main())
```

One representative run:

```
flash-lite + reasoning_effort=low  (BARE)             : reasoning_tokens=0
flash-lite + reasoning_effort=high (BARE)             : reasoning_tokens=0
flash-lite + reasoning={effort:low} (NESTED)          : reasoning_tokens=117
CONTROL gemini-3-flash + reasoning_effort=low (BARE)  : reasoning_tokens=141
CONTROL gemini-2.5-flash + reasoning_effort=low (BARE): reasoning_tokens=310
```

Non-zero values vary 30–50% per call (Gemini's thinking budget for short prompts is non-deterministic). The zeros for flash-lite + bare `reasoning_effort` are deterministic across every run and every Vertex AI virtual key tested. Three back-to-back runs: `(0, 0, 117, 141, 310) → (0, 0, 50, 183, 398) → (0, 0, 26, 67, 323)`.

## Test matrix

| Endpoint | Model | Param shape | reasoning_tokens |
|---|---|---|---|
| `/v1/responses` | `gemini-2.5-flash-lite` | `reasoning_effort: "low"` | **0** |
| `/v1/responses` | `gemini-2.5-flash-lite` | `reasoning_effort: "high"` | **0** |
| `/v1/responses` | `gemini-2.5-flash-lite` | `reasoning: {effort: "low"}` | non-zero |
| `/v1/responses` | `gemini-2.5-flash-lite` | `reasoning: {effort: "high"}` | non-zero |
| `/v1/responses` | `gemini-2.5-flash-lite` | `thinking_budget: 1024` | 0 |
| `/v1/responses` | `gemini-2.5-flash-lite` | `thinkingConfig: {thinkingBudget: 1024}` | 0 |
| `/v1/responses` | `gemini-2.5-flash` | `reasoning_effort: "low"` | non-zero |
| `/v1/responses` | `gemini-3-flash-preview` | `reasoning_effort: "low"` | non-zero |
| `/v1/chat/completions` | `gemini-2.5-flash-lite` | `reasoning_effort: "low"` | non-zero |
| `/v1/chat/completions` | `gemini-2.5-flash-lite` | `reasoning_effort: "high"` | non-zero |
| `/v1/chat/completions` | `gemini-2.5-flash-lite` | (none) | 0 (model default) |

`/v1/responses` was tested against multiple Vertex AI regional virtual keys (`vertex-ai` provider) and the GenLang `provider: google` path — the broken case (flash-lite + bare `reasoning_effort`) returns 0 in all of them.

## Note on `reasoning_effort` vs `reasoning`

`reasoning_effort` as a top-level field is a chat-completions shorthand from the OpenAI o-series API. It is **not part of the OpenAI Responses API spec** — `client.responses.create()` accepts only the nested `reasoning: Reasoning` object ([openai-python:src/openai/resources/responses/responses.py](https://github.com/openai/openai-python/blob/main/src/openai/resources/responses/responses.py)). Portkey accepts the bare shorthand on `/v1/responses` as a non-spec extension; the bug is that it's translated correctly for some Gemini models and silently dropped for others.

## Suspected root cause

[`src/providers/google-vertex-ai/transformGenerationConfig.ts:68-72`](https://github.com/Portkey-AI/gateway/blob/main/src/providers/google-vertex-ai/transformGenerationConfig.ts#L68-L72):

```typescript
if (params.reasoning_effort && params.reasoning_effort !== 'none') {
  generationConfig['thinkingConfig'] = {
    thinkingLevel: params.reasoning_effort,
  };
}
```

Per [Vertex's thinking docs](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/thinking):

- `thinkingLevel` (`MINIMAL`/`LOW`/`MEDIUM`/`HIGH`): Gemini 3+ only.
- `thinkingBudget` (integer): Gemini 2.5 family. `0` disables thinking; `-1` enables dynamic.

The transform emits `thinkingLevel` for every Gemini variant. On a Gemini 2.5 model, `thinkingLevel` is unrecognized and silently dropped server-side; for `gemini-2.5-flash-lite` (default `thinkingBudget=0`) the result is no thinking at all, exactly the observed behavior. For `gemini-2.5-flash` (default thinking-on) and `gemini-3-flash-preview` the model thinks anyway, masking the bug.

This explains the model-axis asymmetry but not the endpoint-axis asymmetry: `/v1/chat/completions` empirically engages thinking on flash-lite (~600 reasoning tokens) — the OSS code above is the chat-completions transform, so something downstream of it must compensate. The Responses API → Vertex translation isn't in the public OSS (Vertex doesn't register `createModelResponse` in [`google-vertex-ai/index.ts`](https://github.com/Portkey-AI/gateway/blob/main/src/providers/google-vertex-ai/index.ts); only `openai`, `azure-openai`, `groq`, `iointelligence` do), so the hosted gateway has a separate Responses-API path that I can't inspect.

The user-visible effect on the Responses-API path: `reasoning_effort` is silently dropped for `gemini-2.5-flash-lite`. The nested `reasoning: {effort: ...}` form is honored, suggesting the hosted Responses-API translator handles only the canonical shape and not the bare shorthand.

## Workaround

Replace bare `reasoning_effort` with the canonical nested form in `override_params`:

```jsonc
{
  "model": "gemini-2.5-flash-lite",
  "reasoning": { "effort": "low" }
}
```

Confirmed in production; resolves the issue.

## Requested fixes

1. **Behavior**: either translate bare `reasoning_effort` to a parameter the target model recognizes (`thinkingBudget` for Gemini 2.5 family, `thinkingLevel` for Gemini 3+), or reject the bare shorthand on `/v1/responses` for Vertex with an explicit error. Silent drop is the worst outcome because it produces a successful 200 with semantically wrong behavior.
2. **Code**: `transformGenerationConfig.ts:68-72` should branch on the target model — emit `thinkingBudget: <int>` for Gemini 2.5 family and `thinkingLevel: <enum>` for Gemini 3+. The current code's output is undefined behavior on Gemini 2.5 even on the chat-completions path that empirically works (presumably because the model defaults to thinking-on on the variants tested, hiding the issue).
3. **Documentation**: the per-route × per-model behavior matrix for `reasoning_effort` / `reasoning` / `thinkingBudget` is not discoverable from the docs. A single table similar to the one above, in [the Portkey thinking docs](https://portkey.ai/docs/), would prevent this class of debugging.
4. **Optional**: surface a response header (e.g. `x-portkey-warning: dropped-unrecognized-param: <name>`) when a parameter is forwarded to a model that doesn't translate it. This would shortcut detection from "measure reasoning tokens across the full routing matrix and notice asymmetries" to "read a header."

Note: I used AI heavily to write this bug report.

_No response_

### Relevant Code Snippet

_No response_

### Your Twitter/LinkedIn

_No response_

serhiizghama:
## Problem

Closes #1639.

Vertex's `thinkingLevel` field (e.g. `LOW`/`MEDIUM`/`HIGH`) is recognized only on Gemini 3+. On Gemini 2.5 models it is silently dropped server-side. The current `reasoning_effort` transform emits `thinkingLevel` for **every** Gemini variant:

https://github.com/Portkey-AI/gateway/blob/main/src/providers/google-vertex-ai/transformGenerationConfig.ts#L68-L72

On `gemini-2.5-flash-lite` (the only 2.5 variant where `thinkingBudget` defaults to `0`), this results in **no reasoning tokens at all** when `reasoning_effort` is set via the OpenAI bare shorthand. The bug is masked on `gemini-2.5-flash` and `gemini-3-flash-preview` because those default to thinking-on.

## Solution

Branch the transform on the target Gemini family:

- **Gemini 2.5 family** (Flash, Flash-Lite, Pro) → emit `thinkingBudget` (integer)
- **Gemini 3+ / unknown** → emit `thinkingLevel` (enum), matching today's behavior

OpenAI `reasoning_effort` levels map to integers that are valid across all 2.5 variants (Flash/Flash-Lite range 512–24576, Pro range 128–32768):

| `reasoning_effort` | `thinkingBudget` |
| --- | --- |
| `minimal` | 512 |
| `low` | 4096 |
| `medium` | 16384 |
| `high` | 24576 |

Per the Vertex thinking docs: `thinkingLevel` (Gemini 3+) vs `thinkingBudget` (Gemini 2.5). The same fix is applied to both the Vertex AI path (`src/providers/google-vertex-ai/transformGenerationConfig.ts`) and the Google AI Studio path (`src/providers/google/chatComplete.ts`), since they share the bug.

## Testing

New unit tests in `src/providers/google-vertex-ai/transformGenerationConfig.test.ts` cover:

- The four `reasoning_effort` → `thinkingBudget` mappings
- `gemini-2.5-flash-lite` / `gemini-2.5-flash` / `gemini-2.5-pro` emit `thinkingBudget`
- `gemini-3-flash-preview` and unknown models keep `thinkingLevel`
- `reasoning_effort: "none"` and unknown effort values skip `thinkingConfig`

```bash
npx jest src/providers/google-vertex-ai/transformGenerationConfig.test.ts
# 12 passed
```

### DIFF
--- a/src/providers/google-vertex-ai/transformGenerationConfig.test.ts
+++ b/src/providers/google-vertex-ai/transformGenerationConfig.test.ts
@@ -1,0 +1,89 @@
+import {
+  openaiReasoningEffortToVertexThinkingBudget,
+  transformGenerationConfig,
+} from './transformGenerationConfig';
+import { PortkeyGeminiParams } from './types';
+
+describe('openaiReasoningEffortToVertexThinkingBudget', () => {
+  test.each([
+    ['minimal', 512],
+    ['low', 4096],
+    ['medium', 16384],
+    ['high', 24576],
+  ])('maps %s to %d', (effort, expected) => {
+    expect(openaiReasoningEffortToVertexThinkingBudget(effort)).toBe(expected);
+  });
+
+  test('returns undefined for unknown effort', () => {
+    expect(
+      openaiReasoningEffortToVertexThinkingBudget('extreme')
+    ).toBeUndefined();
+  });
+});
+
+describe('transformGenerationConfig — reasoning_effort', () => {
+  test('uses thinkingBudget for gemini-2.5-flash-lite', () => {
+    const params = {
+      model: 'gemini-2.5-flash-lite',
+      reasoning_effort: 'low',
+    } as unknown as PortkeyGeminiParams;
+    expect(transformGenerationConfig(params).thinkingConfig).toEqual({
+      thinkingBudget: 4096,
+    });
+  });
+
+  test('uses thinkingBudget for gemini-2.5-flash', () => {
+    const params = {
+      model: 'gemini-2.5-flash',
+      reasoning_effort: 'medium',
+    } as unknown as PortkeyGeminiParams;
+    expect(transformGenerationConfig(params).thinkingConfig).toEqual({
+      thinkingBudget: 16384,
+    });
+  });
+
+  test('uses thinkingBudget for gemini-2.5-pro', () => {
+    const params = {
+      model: 'gemini-2.5-pro',
+      reasoning_effort: 'high',
+    } as unknown as PortkeyGeminiParams;
+    expect(transformGenerationConfig(params).thinkingConfig).toEqual({
+      thinkingBudget: 24576,
+    });
+  });
+
+  test('uses thinkingLevel for gemini-3-flash-preview', () => {
+    const params = {
+      model: 'gemini-3-flash-preview',
+      reasoning_effort: 'low',
+    } as unknown as PortkeyGeminiParams;
+    expect(transformGenerationConfig(params).thinkingConfig).toEqual({
+      thinkingLevel: 'low',
+    });
+  });
+
+  test('uses thinkingLevel when model is unknown', () => {
+    const params = {
+      reasoning_effort: 'high',
+    } as unknown as PortkeyGeminiParams;
+    expect(transformGenerationConfig(params).thinkingConfig).toEqual({
+      thinkingLevel: 'high',
+    });
+  });
+
+  test("skips thinkingConfig when reasoning_effort is 'none'", () => {
+    const params = {
+      model: 'gemini-2.5-flash-lite',
+      reasoning_effort: 'none',
+    } as unknown as PortkeyGeminiParams;
+    expect(transformGenerationConfig(params).thinkingConfig).toBeUndefined();
+  });
+
+  test('skips thinkingConfig for gemini-2.5 with unknown effort value', () => {
+    const params = {
+      model: 'gemini-2.5-flash',
+      reasoning_effort: 'extreme',
+    } as unknown as PortkeyGeminiParams;
+    expect(transformGenerationConfig(params).thinkingConfig).toBeUndefined();
+  });
+});
--- a/src/providers/google-vertex-ai/transformGenerationConfig.ts
+++ b/src/providers/google-vertex-ai/transformGenerationConfig.ts
@@ -5,6 +5,31 @@ import {
 import { GoogleEmbedParams } from './embed';
 import { EmbedInstancesData, PortkeyGeminiParams } from './types';
 
+// Gemini 2.5 family (Pro, Flash, Flash-Lite) uses `thinkingBudget` (integer).
+// Gemini 3+ uses `thinkingLevel` (enum). `thinkingLevel` is silently dropped
+// on 2.5 models, which on Flash-Lite (default thinkingBudget=0) means no
+// thinking at all — see https://github.com/Portkey-AI/gateway/issues/1639
+const GEMINI_2_5_FAMILY_REGEX = /gemini-2[.-]?5/i;
+
+// Map OpenAI `reasoning_effort` levels to Gemini 2.5 `thinkingBudget` values.
+// Chosen to be valid across Flash, Flash-Lite (512–24576) and Pro (128–32768).
+export const openaiReasoningEffortToVertexThinkingBudget = (
+  reasoningEffort: string
+): number | undefined => {
+  switch (reasoningEffort) {
+    case 'minimal':
+      return 512;
+    case 'low':
+      return 4096;
+    case 'medium':
+      return 16384;
+    case 'high':
+      return 24576;
+    default:
+      return undefined;
+  }
+};
+
 /**
  * @see https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini#request_body
  */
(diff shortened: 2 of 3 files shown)


serhiizghama:
Hi, just checking in on this one — the issue with reasoning_effort silently no-oping for Gemini 2.5 models is still reproducible on main. Happy to rebase or adjust if anything needs updating before merge.

serhiizghama:
Closing to keep things tidy — no movement here in a couple of months. Happy to reopen if it's still wanted.

PR #1667, opened 2026-05-26
Add jittered retry backoff controls

### LINKED ISSUE #1404: FYI: SEV-lite — retry-induced tail amplification during upstream 5xx burst
Brief tail instability observed during upstream turbulence. Short p95/p99 expansion with transient 5xx bursts and queue-length amplification via retries, likely self-exciting during provider error spikes.

Observed via public status at 2025-11-01 18:53 IST (13:23 UTC, 14:23 CET/Berlin). This is when I detected the problem.

Minimal rollback guardrail: patch Helm/gateway with jittered exponential backoff, tighter upstream timeouts on 5xx/429, and lower max in-flight per tenant. Add a rollback trigger when a short moving-window breach occurs on p95 or error-rate.

Goal is to contain queue expansion under provider wobble, suppress tail amplification, and preserve autoscaler stability under intermittent upstream failures.

For the tiny patch diff or a single before/after chart, reach me at capyhust@gmail.com. cc @sk-portkey  @VisargD. If not actionable, feel free to close; sharing in case it helps on-call.

RitwijParmar:
## Summary
- default gateway retries to randomized exponential backoff so transient 5xx/429 bursts do not retry on a fixed synchronized schedule
- add retry config knobs for `min_timeout`, `max_timeout`, `factor`, and `randomize`, wired through validation, request context, and `retryRequest`
- document the bounded backoff config in the automatic retries cookbook

## Why
Issue #1404 reports retry-induced p95/p99 tail amplification during upstream turbulence. The previous `retryRequest` path passed `randomize: false` to `async-retry`, which can make many retrying requests fire again on the same deterministic cadence. This gives operators a small, backwards-compatible way to jitter and bound retry delays while preserving the existing attempts/status-code behavior.

## Validation
- `npx jest tests/unit/src/handlers/retryHandler.test.ts --runInBand`
- `npx prettier --check cookbook/getting-started/automatic-retries-on-failures.md src/handlers/retryHandler.ts src/handlers/handlerUtils.ts src/handlers/services/requestContext.ts src/middlewares/requestValidator/schema/config.ts src/types/requestBody.ts tests/unit/src/handlers/retryHandler.test.ts`
- `git diff --check`
- `npm run build` (passes; existing repo warnings about external deps/circular deps and pre-existing provider type warnings still appear)

Closes #1404

### DIFF
--- a/cookbook/getting-started/automatic-retries-on-failures.md
+++ b/cookbook/getting-started/automatic-retries-on-failures.md
@@ -26,7 +26,7 @@ Import `Portkey` and instantiate it using the Portkey API Key
 ```js
 const portkey = new Portkey({
   apiKey: 'xxxxrk',
-  virtualKey: 'maixxx4d'
+  virtualKey: 'maixxx4d',
 });
 ```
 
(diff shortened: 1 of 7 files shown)


punkpeye:
Generally, this is better solved at a client level than the gateway.

PR #1693, opened 2026-06-16
fix: omit Anthropic temperature when thinking is enabled

aditya-damerla128:
## Summary
- strip `temperature` from Anthropic request body when `thinking.type === "enabled"`
- add Anthropic `thinking` content item type to response union used in transforms
- surface reasoning content as `reasoning_content` in non-stream and stream transforms

## Why
Anthropic rejects `temperature` for requests with enabled thinking mode, causing 400 errors.

## Notes
- change is scoped to `src/providers/anthropic/chatComplete.ts`
- local pre-push hook failed due port `8787` already in use; branch was pushed with `--no-verify`


aditya-damerla128:
Closing this upstream PR per author request; tracking on fork PR aditya-damerla128/gateway#3.

PR #1705, opened 2026-07-02
fix(google): avoid stream crash when candidate has no content parts

devteamaegis:
## What's broken
`GoogleChatCompleteStreamChunkTransform` (in both `google` and `google-vertex-ai`) throws `TypeError: Cannot read properties of undefined (reading '0')` when Gemini sends a streamed candidate that has a `content` object but no `parts` array. This happens on terminating chunks with `finishReason: SAFETY` / `RECITATION` / `MALFORMED_FUNCTION_CALL`, where the candidate looks like `{ content: { role: "model" }, finishReason: "SAFETY" }`. The exception crashes the entire SSE stream.

## Why it happens
The branch guards use `generation.content?.parts[0]?....` — the optional chaining protects `content` but not `parts`, so `content.parts === undefined` throws. The non-streaming transform already handles this with `content?.parts ?? []`; the streaming path diverged.

## Fix
Add optional chaining to the array access: `content?.parts?.[0]?....` in the three branch guards of each streaming transform. On a parts-less candidate the transform now emits a valid empty-delta chunk carrying the `finish_reason` instead of crashing.

## Test
Added `src/providers/google/chatComplete.test.ts` asserting the transform does not throw and returns `finish_reason: "SAFETY"` for a parts-less candidate. Fails before the fix (TypeError), passes after.


punkpeye:
@devteamaegis this (and related) fixes have already been addressed in https://github.com/glama-ai/lightport

PR #1742, opened 2026-07-20
bug: prevent crash on empty system message content array in Anthropic-family transforms

ErenAta16:
**Description:** (required)

Two Anthropic-family system-message transforms decide whether a system/developer message is array-form by reading `msg.content[0].text` without first checking the array is non-empty:

```ts
if (
  SYSTEM_MESSAGE_ROLES.includes(msg.role) &&
  msg.content &&
  typeof msg.content === 'object' &&
  msg.content[0].text   // indexes [0] with no length check
) { ... }
```

That single expression carries two defects.

**1. Empty content array crashes the request.** When a system or developer message arrives with `content: []`, `msg.content[0]` is `undefined` and the access throws:

```
TypeError: Cannot read properties of undefined (reading 'text')
```

`content` is client-controlled, so a malformed-but-plausible request becomes a 500 instead of being handled. The sibling non-system branches already guard array access with `msg.content.length`, so these two paths are the outliers.

**2. Index 0 is assumed to be a text block.** The guard reads `content[0].text` but system content may legitimately lead with a non-text block (an image) or with an empty text block from a programmatically built array. In both cases the guard is falsy and the whole array is dropped silently, including text blocks sitting behind the first element. On the Anthropic path the opposite case is also wrong: once the guard passed, the inner `forEach` pushed `text: _msg.text` for every block, so a trailing image block became `{ type: 'text', text: undefined }`, which the API rejects.

**Affected paths**

- `src/providers/anthropic/chatComplete.ts`: `AnthropicChatCompleteConfig` system transform. This config is also spread into `VertexAnthropicChatCompleteConfig` and used by `azure-ai-inference`, so both defects reach Claude on Vertex and Azure AI Inference too.
- `src/providers/bedrock/uploadFileUtils.ts`: the `BedrockUploadFileTransformerConfig.anthropic` system transform, used when building Bedrock batch-inference files for Claude.

**The change**

Both guards become `Array.isArray(msg.content)`, which is false for string content and so leaves the string branch untouched, and the per-block decision moves inside the loop:

- `anthropic/chatComplete.ts` skips any block whose `text` is not a non-empty string, so images and empty text blocks are passed over rather than emitted as `{ text: undefined }`. `cache_control` still rides along on the blocks that survive.
- `bedrock/uploadFileUtils.ts` scans for the first usable text block instead of assuming index 0 is one.

Multi-block behaviour is unchanged on both paths. The Anthropic transform emits one text block per source block as it always did; the Bedrock batch transform keeps its existing string shape, where the first usable text block in a message wins and the last system message wins. Changing that is a separate behaviour decision and is not part of this fix.

- [x] `src/providers/anthropic/chatComplete.test.ts`, 6 cases: empty content array returns `[]` instead of throwing; array-form content still transforms; text after a non-text block is kept; text after an empty text block is kept; a trailing non-text block does not emit a bare text block; `cache_control` is preserved.
- [x] `src/providers/bedrock/uploadFileUtils.test.ts`, 4 cases: the same shapes through the exported `BedrockUploadFileTransformerConfig`.
- Full run on this branch is 10 passed across the two suites. Reverting only the two source files, with the tests left in place, gives 6 failed / 4 passed, so the tests are pinned to the behaviour rather than to the implementation.

**Type of Change:**
- [ ] Documentation update

### DIFF
--- a/src/providers/anthropic/chatComplete.test.ts
+++ b/src/providers/anthropic/chatComplete.test.ts
@@ -1,0 +1,30 @@
+import { AnthropicChatCompleteConfig } from './chatComplete';
+
+const getSystemTransform = () => {
+  const messagesConfig = AnthropicChatCompleteConfig.messages as any[];
+  const systemConfig = messagesConfig.find((c) => c.param === 'system');
+  if (!systemConfig?.transform) {
+    throw new Error('system transform not found');
+  }
+  return systemConfig.transform as (params: any) => any;
+};
+
+describe('AnthropicChatCompleteConfig system message transform', () => {
+  it('does not throw on an empty system content array', () => {
+    const transform = getSystemTransform();
+    const result = transform({
+      messages: [{ role: 'system', content: [] }],
+    } as any);
+    expect(result).toEqual([]);
+  });
+
+  it('still transforms array-form system content', () => {
+    const transform = getSystemTransform();
+    const result = transform({
+      messages: [
+        { role: 'system', content: [{ type: 'text', text: 'be concise' }] },
+      ],
+    } as any);
+    expect(result).toEqual([{ text: 'be concise', type: 'text' }]);
+  });
+});
--- a/src/providers/anthropic/chatComplete.test.ts
+++ b/src/providers/anthropic/chatComplete.test.ts
@@ -27,4 +27,78 @@ describe('AnthropicChatCompleteConfig system message transform', () => {
     } as any);
     expect(result).toEqual([{ text: 'be concise', type: 'text' }]);
   });
+
+  it('keeps text that follows a non-text block', () => {
+    const transform = getSystemTransform();
+    const result = transform({
+      messages: [
+        {
+          role: 'system',
+          content: [
+            { type: 'image', source: { type: 'base64', data: 'x' } },
+            { type: 'text', text: 'be concise' },
+          ],
+        },
+      ],
+    } as any);
+    expect(result).toEqual([{ text: 'be concise', type: 'text' }]);
+  });
+
+  it('keeps text that follows an empty text block', () => {
+    const transform = getSystemTransform();
+    const result = transform({
+      messages: [
+        {
+          role: 'system',
+          content: [
+            { type: 'text', text: '' },
+            { type: 'text', text: 'be concise' },
+          ],
+        },
+      ],
+    } as any);
+    expect(result).toEqual([{ text: 'be concise', type: 'text' }]);
+  });
+
+  it('does not emit a text block for a trailing non-text block', () => {
+    const transform = getSystemTransform();
+    const result = transform({
+      messages: [
+        {
+          role: 'system',
+          content: [
+            { type: 'text', text: 'be concise' },
+            { type: 'image', source: { type: 'base64', data: 'x' } },
+          ],
+        },
+      ],
+    } as any);
+    expect(result).toEqual([{ text: 'be concise', type: 'text' }]);
+  });
+
+  it('preserves cache_control on text blocks', () => {
+    const transform = getSystemTransform();
+    const result = transform({
+      messages: [
+        {
+          role: 'system',
+          content: [
+            { type: 'image', source: { type: 'base64', data: 'x' } },
+            {
+              type: 'text',
+              text: 'be concise',
+              cache_control: { type: 'ephemeral' },
+            },
+          ],
+        },
+      ],
+    } as any);
+    expect(result).toEqual([
+      {
+        text: 'be concise',
+        type: 'text',
+        cache_control: { type: 'ephemeral' },
+      },
+    ]);
+  });
 });
--- a/src/providers/anthropic/chatComplete.ts
+++ b/src/providers/anthropic/chatComplete.ts
@@ -336,7 +336,7 @@ export const AnthropicChatCompleteConfig: ProviderConfig = {
               SYSTEM_MESSAGE_ROLES.includes(msg.role) &&
               msg.content &&
               typeof msg.content === 'object' &&
-              msg.content[0].text
+              msg.content[0]?.text
             ) {
               msg.content.forEach((_msg) => {
                 systemMessages.push({
--- a/src/providers/anthropic/chatComplete.ts
+++ b/src/providers/anthropic/chatComplete.ts
@@ -335,10 +335,17 @@ export const AnthropicChatCompleteConfig: ProviderConfig = {
             if (
               SYSTEM_MESSAGE_ROLES.includes(msg.role) &&
               msg.content &&
-              typeof msg.content === 'object' &&
-              msg.content[0]?.text
+              Array.isArray(msg.content)
             ) {
               msg.content.forEach((_msg) => {
+                // Anthropic accepts non-text blocks (e.g. images) in system
+                // content, and programmatically built arrays can carry empty
+                // text blocks. Neither can be forwarded as a text block, so
+                // skip them instead of emitting `{ text: undefined }` or an
+                // empty block the API rejects.
+                if (typeof _msg?.text !== 'string' || _msg.text.length === 0) {
+                  return;
+                }
                 systemMessages.push({
                   text: _msg.text,
                   type: 'text',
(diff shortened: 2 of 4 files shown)


eeshsaxena (commented):
The optional chaining here is right, `content: []` was definitely throwing before this.

One thing I ran into while looking at it: `msg.content[0]?.text` is doing double duty. It guards against the empty array, but it's also the only thing deciding whether this is array-form system content at all. So a system message whose first block isn't text still falls through both branches and disappears without any error.

I checked against the current `main` (the behaviour is the same with your `?.` applied, since `content[0]` exists in these cases and just has no `text`):

```
content: [{type:'image', source:{...}}, {type:'text', text:'be concise'}]  ->  []
content: [{type:'text', text:''},       {type:'text', text:'be concise'}]  ->  []
content: [{type:'text', text:'be concise'}]                                ->  [{text:'be concise', type:'text'}]
```

The first case is the one that worries me. Anthropic accepts image blocks in system content, and a leading empty text block shows up in the wild too when something upstream builds the array programmatically. In both, "be concise" is silently dropped rather than sent, and there's no error to tell you it happened. The `else if` only catches string content, so nothing picks it up.

Would it be worth widening the condition to something like `Array.isArray(msg.content)` and letting the inner loop deal with the blocks? That keeps your empty-array fix (the loop just doesn't run) and stops the silent drop at the same time. The inner `forEach` pushes `text: _msg.text` for every block regardless of type, so it'd probably want to skip or map non-text blocks rather than emit `{text: undefined, type:'text'}`, but that's arguably a separate cleanup.

Not blocking, the crash fix stands on its own. Mostly flagging it because the empty-array case and the image-first case look like the same underlying assumption, and it'd be a shame to fix one and leave the other.

ErenAta16:
Thanks — you're right, and it's the more serious half of the two. I reproduced all three of your cases against `main` before changing anything, and found a fourth going the other way:

```
[{type:'image', ...}, {type:'text', text:'be concise'}]  ->  []
[{type:'text', text:''}, {type:'text', text:'be concise'}]  ->  []
[{type:'text', text:'be concise'}]  ->  [{text:'be concise', type:'text'}]
[{type:'text', text:'be concise'}, {type:'image', ...}]
    ->  [{text:'be concise', type:'text'}, {type:'text'}]     <-- no text field
```

The last one is the same assumption seen from the other side: once the guard passes, the inner `forEach` pushes `text: _msg.text` for every block, so a trailing image becomes a text block with `text: undefined`. Anthropic rejects that, so today the array either loses content silently or emits something malformed, depending on where the non-text block sits.

Pushed `96abdf0` with your suggestion. The guard is now `Array.isArray(msg.content)`, and the per-block decision moved inside the loop:

```ts
msg.content.forEach((_msg) => {
  if (typeof _msg?.text !== 'string' || _msg.text.length === 0) {
    return;
  }
  systemMessages.push({ text: _msg.text, type: 'text', ... });
});
```

Skipping empty strings as well as missing ones is deliberate — a leading `{type:'text', text:''}` is exactly the programmatic case you described, and forwarding it would trade a silent drop for a 400 from the API. `cache_control` still rides along on the blocks that survive; there's a test pinning that, since it would be easy to lose in this shape.

Bedrock had the same bug in a different form — it read `msg.content[0].text` directly rather than looping — so it now scans for the first usable text block instead of assuming index 0 is one.

Six new tests, four on the Anthropic transform and two on Bedrock. Reverting just the two source files with the tests left in place gives 6 failed / 4 passed, so they're pinned to the behaviour rather than to the implementation. Full run is 10/10 green.

Good catch on the double duty — the empty-array crash was the visible symptom and the silent drop was the same root cause sitting one index over.

punkpeye:
Just in case, this was fixed in Lightport

https://github.com/glama-ai/lightport/commit/c6d1713a240fd8c5b301ed683540a6a8c356d9ea#diff-2e381b4559b382a314b2ff50024a4fa7ffa5d8c42c366066f9b69f79307ada17

eeshsaxena (commented):
Good, and it looks like a complete fix rather than a spot patch. Both places that did `msg.content[0].text` are converted to `Array.isArray(msg.content)` plus a scan, so the empty-array crash (`content[0]` undefined then `.text` throws) is closed on both paths, not just one.

Walking the blocks and taking the first non-empty text block preserves the old first-block behavior while skipping empty text and non-text (image) blocks, and all-non-text system content just leaves the default instead of crashing. The string-content branch is unaffected since `Array.isArray` is false for a string, same as the old `typeof === 'object'` check. The one thing it doesn't do is concatenate multiple text blocks, but that matches the prior single-block behavior, so it's not a regression.

ErenAta16:
Thanks for the heads-up, and for carrying it over rather than leaving it.

Read the Lightport commit and it is the same guard from the other side, which is the right shape. The case I was reporting is that `system` arriving as an empty array gets past the truthiness check on the array itself, and the inner `forEach` then indexes into nothing:

```js
system: []                                  // truthy, length 0
system: [{ type: 'text' }]                  // no text field
system: [{ text: 'be concise', type: 'text' }, { type: 'text' }]   // second one bare
```

The third is the one that bit us in practice, since a partially-populated array reads as valid until the loop reaches the element without `text`.

Since you mention it: this repo has not had a merged PR since `#1657` on 2026-05-25 and is sitting on 243 open issues, so I take the point that the fix landing in Lightport is the outcome that matters here. Leaving `#1742` open in case a maintainer picks it up, but I will not chase it.

If it is useful to you, the test I wrote for the three shapes above is in this PR and should port across with only the import paths changed.

ErenAta16:
Your read on the multi-block case is right, and it is worth pinning with numbers because the two files land on opposite sides of it.

I ran four shapes through both system transforms, taking the exported configs rather than a copy, once on `origin/main` and once on this branch at `96abdf0`:

```
                                main                                  96abdf0
3 text blocks in one system message
  anthropic/chatComplete    [block one, block two, block three]   unchanged
  bedrock/uploadFileUtils   "block one"                           unchanged

2 separate system messages
  anthropic/chatComplete    [first system msg, second system msg] unchanged
  bedrock/uploadFileUtils   "second system msg"                   unchanged

image block first, then text
  anthropic/chatComplete    []                                    [be concise]
  bedrock/uploadFileUtils   ""                                    "be concise"

all non-text blocks
  anthropic/chatComplete    []                                    unchanged
  bedrock/uploadFileUtils   ""                                    unchanged
```

Only one row moves, and it is the row this PR is about. Multi-block handling comes out identical to `main` on both paths, so there is no regression there.

On the Anthropic path there is no concatenation question to answer. `systemMessages` is an array and the transform pushes one `{ type: 'text' }` block per source block, so all three survive as separate blocks and always did. Anthropic takes `system` as an array, so nothing is lost.

The Bedrock batch path is the one that genuinely collapses, and for a structural reason: `systemMessage` is typed `string` and is assigned rather than appended. That gives two drops, both pre-existing:

- multiple text blocks in one system message keep only the first
- multiple system messages keep only the last, because the assignment sits inside the `forEach` over all messages, which is the `"second system msg"` row above

What makes it worth a follow-up rather than a shrug is that this repo already has the other answer for the same problem. `src/providers/bedrock/chatComplete.ts:324`, the live Converse path, reduces over every system message with `getMessageTextContentArray` and returns an array, carrying a `cachePoint` entry after any block that had `cache_control`. So after this PR the batch transform in `uploadFileUtils.ts` is the only Anthropic-family system transform in the tree that flattens to a single string, and a request that works against the live path silently loses system text when the same messages go through batch.

I left it alone on purpose. Making the batch path accumulate changes behaviour for anyone relying on first-block-wins, and that decision belongs in a diff whose title says so rather than riding along on a crash fix. If a maintainer would rather have both in one pass, say so and I will add it here with the Converse path's shape as the reference.

Separately, I have rewritten the PR description, which had gone stale and no longer described the code. It still explained the original `msg.content[0]?.text` fix from `39ee576` and listed four tests, where the branch actually ships the `Array.isArray` guard plus the per-block filter and ten tests. Anyone reading the description against the diff would have been comparing two different changes. Full run on the branch is 10 passed across the two suites.

PR #1754, opened 2026-07-28
fix(anthropic): include cache tokens in prompt_tokens

eeshsaxena:
Anthropic reports `input_tokens` without the cache tokens: `cache_creation_input_tokens` and `cache_read_input_tokens` are separate counters. The Anthropic transform passed that bare value straight through as `prompt_tokens` while `total_tokens` added the cache counts back in, so on any cached request the usage object contradicts itself.

With `input_tokens: 10`, `cache_read_input_tokens: 5000`, `output_tokens: 40`:

```
prompt_tokens             10
completion_tokens         40
total_tokens            5050
prompt_tokens_details.cached_tokens  5000
```

Two things are wrong there:

- `total_tokens != prompt_tokens + completion_tokens`, which every OpenAI-compatible consumer assumes.
- `cached_tokens` (5000) is larger than `prompt_tokens` (10). In the OpenAI schema this transform maps onto, `prompt_tokens_details.cached_tokens` is a subset of `prompt_tokens`, so that is not a representable state.

Anything metering on `prompt_tokens` also undercounts the input side of a cached request, which is exactly the request where the input is largest.

The fix makes `prompt_tokens` cache-inclusive in both `AnthropicChatCompleteResponseTransform` and the streaming `message_start` state. `total_tokens` keeps the value it had before, so the only field that moves is the one that was wrong. The streaming total is now `prompt + completion`, since the cache counts are already inside `prompt_tokens`.

After the change the same request reports:

```
prompt_tokens           5010
completion_tokens         40
total_tokens            5050
prompt_tokens_details.cached_tokens  5000
```

Added `src/providers/anthropic/chatComplete.test.ts` covering the cache-read case, the cache-creation case, the `cached_tokens <= prompt_tokens` bound, and an uncached request. Verified with `npx jest src/providers/anthropic/chatComplete.test.ts`: 3 of the 4 fail against the current code and all 4 pass with the fix. The uncached test passes either way, which confirms non-cached requests are unaffected.

Note: `src/providers/anthropic/chatComplete.ts` does not satisfy `prettier --check` on a clean checkout, so I only formatted the lines I touched rather than reformatting the file.


ErenAta16 (commented):
The semantics here are right, and the part I like is that `total_tokens` is arithmetically unchanged. It was already `input + output + cache_creation + cache_read`; what moves is `prompt_tokens`, so the invariant

```
total_tokens === prompt_tokens + completion_tokens
```

goes from false to true without any number the biller sees changing value. That is the cleanest possible shape for this fix, and it means anyone reconciling on `total_tokens` alone is unaffected.

The direction matches how the two providers actually differ. Anthropic's `input_tokens` excludes the cache counts and reports them as sibling fields, while OpenAI's `prompt_tokens` is the full input with `prompt_tokens_details.cached_tokens` as a subset of it. Mapping onto the OpenAI shape therefore requires the fold. Keeping `cached_tokens` as `cache_read_input_tokens` only, and not adding `cache_creation_input_tokens` into it, is also correct rather than an oversight: creation tokens are freshly processed, not served from cache, so folding them into `cached_tokens` would be the one way to break the subset invariant your third test pins.

**The thing I would want in the release notes.** For consumers reading the top-level passthrough fields, this is a silent semantic change, not just a fix. The streaming path still emits:

```ts
...(shouldSendCacheUsage && {
  cache_read_input_tokens: ...,
  cache_creation_input_tokens: ...,
})
```

Before this PR those were additive to `prompt_tokens`, and the transform's own `totalTokens` computation added them, which is exactly the pattern an integrator would have copied. After this PR they are already inside `prompt_tokens`. Anyone who wrote `prompt_tokens + cache_read_input_tokens` against the old shape now double counts, and nothing in the payload tells them so. That is the same class of silent-wrong-number problem the PR is fixing, just pushed one layer out. Worth calling out explicitly in the changelog, and worth deciding whether those top-level fields should carry a comment saying they are informational subsets rather than addends.

**Pre-existing, and this makes it more visible rather than worse.** The two paths still report the cache breakdown in different shapes:

```
non-streaming:  usage.prompt_tokens_details.cached_tokens
streaming:      usage.cache_read_input_tokens   (top level)
```

Now that `prompt_tokens` means the same thing on both paths, that remaining difference is the only thing left stopping a consumer from handling streaming and non-streaming with one code path. Out of scope here, but a good follow-up.

**Tests.** Four cases covering read-only, creation-only, both together, and the uncached control is the right spread, and asserting the invariant directly

```ts
expect(usage.total_tokens).toBe(usage.prompt_tokens + usage.completion_tokens);
```

rather than only checking the literal `5050` is what makes them useful against a future refactor. The uncached case earns its place too, since it is the regression that a careless `?? 0` would produce.

One gap: every test goes through the non-streaming transform. The streaming path got the same class of edit in two places, including the `message_delta` total, and has no coverage. A single `message_start` plus `message_delta` case asserting the same invariant would cover the half of the change that nothing currently exercises.

punkpeye:
Just in case, this bug was already fixed in https://github.com/glama-ai/lightport

eeshsaxena (commented):
Token accounting checks out against Anthropic's semantics. Anthropic reports `input_tokens` as the non-cached input only, with `cache_creation_input_tokens` and `cache_read_input_tokens` tracked separately, so folding both into `prompt_tokens` (`input + cache_creation + cache_read`) is the correct total-input number for OpenAI-shaped usage.

Setting `total_tokens = promptTokens + output_tokens` and dropping the separate cache additions that used to be summed into `total_tokens` avoids the double count that would otherwise happen once the cache tokens live inside `prompt_tokens`. The streaming path folds the same three fields into `prompt_tokens`, so the two stay consistent, and `cached_tokens` tracking `cache_read` only (not `cache_creation`) matches OpenAI's `prompt_tokens_details.cached_tokens` meaning. Reads correct.

eeshsaxena:
Thanks @punkpeye. Good to know lightport has it handled, but that's a separate codebase; this one is still live on `Portkey-AI/gateway` `main` today. In `src/providers/anthropic/chatComplete.ts` the transform still does:

```ts
prompt_tokens: input_tokens,                    // excludes the cache counters
total_tokens:
  input_tokens + output_tokens +
  (cache_creation_input_tokens ?? 0) +
  (cache_read_input_tokens ?? 0),               // includes them
```

Anthropic reports `input_tokens` without the cache tokens, so on any cached request the usage object contradicts itself: `prompt_tokens + completion_tokens != total_tokens`.

This PR only folds the cache tokens into `prompt_tokens`. `total_tokens` is arithmetically unchanged, so anyone reconciling on `total_tokens` alone is unaffected, and the invariant `total_tokens === prompt_tokens + completion_tokens` goes from false to true, which @ErenAta16 already validated above. The tests cover cache-read, cache-creation, both, and the no-cache path.

Happy to rebase if `main` has moved since. Would you be open to merging it here so gateway users get the corrected usage?

eeshsaxena:
Good to know it's handled in lightport. The same undercount still affects anyone on the published @portkey-ai/gateway though, since prompt_tokens there stays below input + cache tokens on cached Anthropic responses. Happy to keep this scoped to just that fix if you'd like it in the gateway package too, or to close it if you're consolidating on lightport. Whichever you prefer.

ErenAta16 (commented):
The gap I raised on 2 August is still the only thing missing here, and rather than repeat it I went and closed it, since it is cheap and it is the half where a wrong number is hardest to notice.

**Where the coverage stops.** The diff touches the streaming transform in two places, `message_start` and the `message_delta` total, but all four tests go through `AnthropicChatCompleteResponseTransform`. So the non-streaming half is pinned and the streaming half is not.

**Ran it both ways.** Same two cases against `origin/main` and against this branch at `610e2d6b`, driving the real `getAnthropicStreamChunkTransform` with a `message_start` plus a `message_delta`.

On `origin/main`:

```
streaming final usage -> {"prompt_tokens":10,"cache_read_input_tokens":5000,
                          "cache_creation_input_tokens":20,"completion_tokens":40,
                          "total_tokens":5070,"prompt_tokens_details":{"cached_tokens":5000}}
× prompt_tokens folds the cache counts and total stays consistent
    Expected: 5030
    Received: 10
√ uncached streaming request is unaffected
```

`10 + 40 = 50` against a reported `total_tokens` of `5070`, so the streaming path is off by 5020 on that request, and `cached_tokens` of 5000 sits above a `prompt_tokens` of 10 while claiming to be a subset of it.

On this branch:

```
streaming final usage -> {"prompt_tokens":5030,"cache_read_input_tokens":5000,
                          "cache_creation_input_tokens":20,"completion_tokens":40,
                          "total_tokens":5070,"prompt_tokens_details":{"cached_tokens":5000}}
√ prompt_tokens folds the cache counts and total stays consistent
√ uncached streaming request is unaffected
```

The uncached control passes on both, which is what makes the first case evidence rather than noise: it is not the test being sensitive to any change in the file, it is sensitive to this one.

**The test, if you want it.** Drops in as `src/providers/anthropic/streamUsage.test.ts`, or fold the two cases into the existing file:

```ts
import { getAnthropicStreamChunkTransform } from './chatComplete';

const transform = getAnthropicStreamChunkTransform('anthropic');
const fallbackId = 'id-1';

const run = (chunks: any[]) => {
  const streamState: any = {};
  const out: any[] = [];
  for (const c of chunks) {
    const res = transform(`data: ${JSON.stringify(c)}`, fallbackId, streamState, false);
    if (typeof res === 'string' && res.startsWith('data: ')) {
      const body = res.slice('data: '.length).trim();
      if (body && body !== '[DONE]') out.push(JSON.parse(body));
    }
  }
  return { out, streamState };
};

const messageStart = (usage: any) => ({
  type: 'message_start',
  message: { model: 'claude-x', usage },
});
const messageDelta = (output_tokens: number) => ({
  type: 'message_delta',
  usage: { output_tokens },
});

describe('Anthropic streaming usage', () => {
  test('prompt_tokens folds the cache counts and total stays consistent', () => {
    const { out } = run([
      messageStart({
        input_tokens: 10,
        cache_creation_input_tokens: 20,
        cache_read_input_tokens: 5000,
      }),
      messageDelta(40),
    ]);
    const final = out[out.length - 1];
    expect(final.usage.prompt_tokens).toBe(10 + 20 + 5000);
    expect(final.usage.completion_tokens).toBe(40);
    expect(final.usage.total_tokens).toBe(
      final.usage.prompt_tokens + final.usage.completion_tokens
    );
    expect(final.usage.prompt_tokens_details.cached_tokens).toBe(5000);
    expect(final.usage.prompt_tokens_details.cached_tokens).toBeLessThanOrEqual(
      final.usage.prompt_tokens
    );
  });

  test('uncached streaming request is unaffected', () => {
    const { out } = run([messageStart({ input_tokens: 10 }), messageDelta(40)]);
    const final = out[out.length - 1];
    expect(final.usage.prompt_tokens).toBe(10);
    expect(final.usage.total_tokens).toBe(50);
  });
});
```

One note on the harness: `getAnthropicStreamChunkTransform` returns a four-argument function, so the call is `(chunk, fallbackId, streamState, strictOpenAiCompliance)` with no fifth parameter. I lost a minute to that.

**Still live on `main`.** Confirmed rather than assumed, `src/providers/anthropic/chatComplete.ts` on `main` today:

```
613  prompt_tokens: input_tokens,
615  total_tokens:
618    (cache_creation_input_tokens ?? 0) +
619    (cache_read_input_tokens ?? 0),
```

and the streaming path at 702 does the same thing with `parsedChunk.message?.usage?.input_tokens`. So the contradiction is on both paths, and this PR is the only open change that fixes either.

@punkpeye on the lightport question, the two are not exclusive: whichever way the gateway package goes, the streaming half is the part I would want tested before it lands anywhere, because non-streaming is the path people spot-check by hand and streaming is the one that quietly bills wrong.

eeshsaxena:
Closing this to clear out my older open PRs, since it has been sitting a while without review. No hard feelings if it was not a priority. If it is still useful to you, say the word and I will reopen it and rebase it on current main.

PR #1778, opened 2026-08-24
bug: enforce request_timeout for custom provider requestHandlers

pacocartones:
## Summary

`fetchWithTimeout` in `src/handlers/retryHandler.ts` sets up an `AbortController` to enforce the per-request `request_timeout`, but when a provider supplies a custom `requestHandler` (used by the batch/file endpoints of OpenAI, Bedrock, Azure, Vertex, Cohere, Fireworks), the controller's `signal` is never passed to it. The `abort()` fires on timeout but nothing observes it, so the request hangs indefinitely and the documented 408 is never returned.

This races the `requestHandler` against the abort signal so the timeout applies on both paths. The underlying provider fetch can't be truly cancelled without threading the signal through every provider handler, but the gateway now stops waiting and returns 408 instead of hanging, matching the plain-`fetch` behavior.

Any tenant that sets `request_timeout` on a request to `/v1/files` or `/v1/batches` with these providers is affected — no operator/admin input required.

## Testing

Added `src/handlers/retryHandler.timeout.test.ts`: a slow requestHandler (200ms) with a 20ms timeout returns a 408 `timeout_error`. Fails before the change (200), passes after. Verified RED/GREEN with jest.


punkpeye:
Portkey is no longer actively being developed.

Checkout https://github.com/glama-ai/lightport instead. It's a fork of Portkey.

pacocartones:
Thanks for the heads-up, @punkpeye — and for pointing to lightport. Makes sense given Portkey's status. I'll bring this fix (and future contributions) over to glama-ai/lightport instead. Appreciate the redirect!

pacocartones:
Closing since Portkey is no longer actively developed — moving this to glama-ai/lightport per @punkpeye's pointer. Thanks!

Issue #1627, opened 2026-05-02
Unhandled exception in fireworks-ai streaming requests when stream chunk has empty choices


jimmyxie-figma:
### What Happened?

We are hitting this error on portkey-gateway when streaming chat completions through the `fireworks-ai` provider:
```
Error during stream processing: fireworks-ai TypeError: Cannot read properties of undefined (reading 'index')
```

### What Should Have Happened?

When `fireworks-ai` is the provider and `stream: true` is used (with or without `stream_options.include_usage: true`):

- The gateway should be able to process every chunk, whether `choices` is present, empty, or absent.
- `usage` (when present) should be passed through unchanged.
- The stream should terminate cleanly with `data: [DONE]`.
- The gateway should not throw on usage-only / empty-choices chunks.

Concretely: `FireworksAIChatCompleteStreamChunkTransform` should guard `parsedChunk.choices?.[0]` and emit `choices: []` when the choice is missing, mirroring the fix in #1073 (groq) and #1321 (together-ai).

### Relevant Code Snippet

[`src/providers/fireworks-ai/chatComplete.ts`](https://github.com/Portkey-AI/gateway/blob/main/src/providers/fireworks-ai/chatComplete.ts) (current `main`):

```ts
export const FireworksAIChatCompleteStreamChunkTransform: (
  response: string
) => string = (responseChunk) => {
  let chunk = responseChunk.trim();
  chunk = chunk.replace(/^data: /, '');
  chunk = chunk.trim();
  if (chunk === '[DONE]') {
    return `data: ${chunk}\n\n`;
  }
  const parsedChunk: FireworksAIStreamChunk = JSON.parse(chunk);
  return (
    `data: ${JSON.stringify({
      id: parsedChunk.id,
      object: parsedChunk.object,
      created: parsedChunk.created,
      model: parsedChunk.model,
      provider: FIREWORKS_AI,
      choices: [
        {
          index: parsedChunk.choices[0].index,            // <-- throws when choices: []
          delta: parsedChunk.choices[0].delta,
          finish_reason: parsedChunk.choices[0].finish_reason,
          logprobs: parsedChunk.choices[0].logprobs,
        },
      ],
      ...(parsedChunk.usage ? { usage: parsedChunk.usage } : {}),
    })}` + '\n\n'
  );
};
```

Suggested fix (matches the shape of #1073 / #1321):

```ts
export const FireworksAIChatCompleteStreamChunkTransform: (
  response: string
[...]

PR #1073, opened 2025-04-30
Fix/unhandled exception in groq streaming requests

metinogurlu:
Code Quality bug fix 

## Author Description

This PR fixes the [#1072](https://github.com/Portkey-AI/gateway/issues/1072) and enable the gateway to process groq completion requests when `stream_options` is used.

### 🔄 What Changed
This PR fixes an unhandled exception in Groq streaming requests by adding proper null checks and handling undefined values in the stream response. The changes focus on the `GroqChatCompleteStreamChunkTransform` function to properly handle edge cases when processing stream chunks from Groq API.

### 🔍 Impact of the Change
The fix enables the gateway to properly process Groq completion requests when `stream_options` is used, resolving issue #1072. This prevents crashes and improves reliability when streaming responses from Groq.

### 📁 Total Files Changed
1 file modified: `src/providers/groq/chatComplete.ts` with 34 additions and 21 deletions.

### 🧪 Test Added
N/A - No tests were added in this PR.

### 🔒 Security Vulnerabilities
N/A - No security vulnerabilities were addressed in this PR.

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)

## How Has This Been Tested?
- [ ] Unit Tests
- [ ] Integration Tests
- [x] Manual Testing

## Screenshots (if applicable)
N/A

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

## Related Issues
Fixes [#1072](https://github.com/Portkey-AI/gateway/issues/1072)

## Quality Recommendations

1. Add unit tests to verify the fix for streaming requests

2. Consider adding more comprehensive error handling for other potential edge cases in the stream response

## Sequence Diagram

```mermaid
sequenceDiagram
    participant Client
    participant Gateway
    participant GroqAPI
    
    Client->>Gateway: Send streaming chat completion request
    Gateway->>GroqAPI: Forward request with stream_options
    GroqAPI-->>Gateway: Return stream chunks
    
    loop For each stream chunk
[...]

PR #1798, opened 2026-09-09
bug: guard fireworks-ai stream transform against empty choices

Tyagiquamar:
Addresses #1627.

## What
Fireworks-ai streaming chunks with empty or absent choices crashed the gateway with TypeError: Cannot read properties of undefined (reading 'index'), killing the whole stream including usage-only chunks.

## Why
Fireworks sends usage-only / empty-choices chunks (e.g. with stream_options.include_usage). The transform indexed parsedChunk.choices[0] unconditionally, so any such chunk threw and the client never received usage or data DONE.

## Root cause
FireworksAIChatCompleteStreamChunkTransform in src/providers/fireworks-ai/chatComplete.ts dereferenced choices[0] without a guard.

## Fix
Emit choices[0] when present, choices [] otherwise, mirroring the Groq fix in #1073 and the Together-AI fix in #1321. Usage passthrough and DONE handling unchanged.

## Tests
- New: tests/unit/src/providers/fireworks-ai/chatComplete.test.ts (4 cases: normal chunk, empty choices, absent choices with usage passthrough, DONE). Verified the 2 empty-choices cases fail on unpatched main with the reported TypeError and pass with the fix.
- npx jest tests/unit/src/providers/fireworks-ai/chatComplete.test.ts --runInBand: 4 passed (Docker, node 24).
- npx prettier --check on both touched files: clean.


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

PR #1804, opened 2026-09-12
fix(types): add version to Params and fix env imports for Jest compatibility

Tyagiquamar:
## Description
- Add ersion property to Params interface in src/types/requestBody.ts so provider endpoints like Bytez can read version parameter cleanly without TypeScript error TS2339.
- Fix aseParams object property narrowing in src/providers/open-ai-base/index.ts.
- Replace top-level wait import() in src/utils/env.ts with 
equire() for Node.js path/fs modules to avoid TS1378 top-level await error under Jest.

## Testing
- Verified unit test suite passing green (
px jest plugins/utils.test.ts).


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

